PR #11, opened 2022-12-26
Fix the build for `main` branch

ruspl-afed:
Add component modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -440,6 +440,12 @@
     </pluginManagement>
   </build>
   <modules>
+    <module>mylyn.commons</module>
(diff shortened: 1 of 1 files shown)


ruspl-afed:
@BeckerFrank do you know why this PR still shows the diff? I thought the `main` already has modules after your change.

BeckerFrank:
> @BeckerFrank do you know why this PR still shows the diff? I thought the `main` already has modules after your change.

No maybe the diff is what in Eclipse is "Compare with Previous Version" and not "Compare With Working Tree".

ruspl-afed:
Not actual after merging #12

PR #12, opened 2022-12-26
fix for https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/967…

BeckerFrank:
For a unknown reason the local build was OK but  Jenkins fail to build the main branch.
With this I partly revert the previous commit.

### DIFF
--- a/org.eclipse.mylyn-target/mylyn-e4.26.target
+++ b/org.eclipse.mylyn-target/mylyn-e4.26.target
@@ -64,6 +64,36 @@
 			<unit id="org.mockito.mockito-core" version="0.0.0"/>
 			<unit id="org.objenesis" version="0.0.0"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/mylyn/snapshots/nightly/builds/"/>
+			<unit id="org.eclipse.mylyn.builds.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.mylyn.builds.sdk.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.mylyn.builds.development.feature.group" version="0.0.0"/>
+		</location>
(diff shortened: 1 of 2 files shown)


org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -64,6 +64,36 @@
 			<unit id="org.mockito.mockito-core" version="0.0.0"/>
 			<unit id="org.objenesis" version="0.0.0"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
```
ruspl-afed, line 67: We already have all these units in reactor for merged repo. Why do we add circular dependency again?

ruspl-afed (commented):
Please have a look at https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/11 @BeckerFrank 
May be we need to merge it first.

BeckerFrank:
> Please have a look at #11 @BeckerFrank
> May be we need to merge it first.

Yes I add the modules to the pom

org.eclipse.mylyn-target/mylyn-e4.26.target
BeckerFrank: > We already have all these units in reactor for merged repo. Why do we add circular dependency again?

With out the added location my local build was fine but the Jenkins Build 
https://ci.eclipse.org/mylyn/job/github-mylyn/job/main/2/ has the error 
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.mylyn.tests 3.26.0.qualifier
[ERROR]   Missing requirement: org.eclipse.mylyn.tests 3.26.0.qualifier requires 'osgi.bundle; org.eclipse.mylyn.bugzilla.tests 3.26.0' but it could not be found

that I could fix with the lines.

The Question is what is different between my local build and the Jenkins build?

ruspl-afed: Now I had a more closer look. You are right, we still need these circular dependencies for some reason, I'll work to eliminate them later.

ruspl-afed (approved):
Let's move, I'll close my #11

PR #15, opened 2022-12-29
Update MylynConfiguration.setup

BeckerFrank:
With a previous commit I changed Maintainer to Mylyn but did not make the change in this configuration


ruspl-afed (approved):
LGTM

PR #16, opened 2022-12-29
format all source files

BeckerFrank:
(see https://github.com/eclipse-mylyn/org.eclipse.mylyn.commons/issues/18)


ruspl-afed (commented):
It will take time

BeckerFrank:
> It will take time

Yes, no problem.
I select all projects and the execute Source -> Format so there no individual changes.

ruspl-afed (approved):
LGTM

wimjongman:
Should we also run a source cleanup and not only a source format?

Issue #10, opened 2022-12-26
Fix the build for `main` branch


ruspl-afed:
Let's try to fix the build for pre-merged version of mylyn sources

ruspl-afed:
@wimjongman @BeckerFrank  please have a look at build results https://ci.eclipse.org/mylyn/job/github-mylyn/job/PR-11/
It may require further cleanup and tuning of deploy, but in general it looks good to me.
WDYT?

wimjongman:
That was quick!

BeckerFrank:
Is there anything left here or can we close this issue?

ruspl-afed:
We have the only point before we can claim the good shape: restore stable and release build process. But this could go to another ticket and this one can be closed.

wimjongman:
> We have the only point before we can claim the good shape: restore stable and release build process. 

I don't understand what you mean by that. Can you explain?

wimjongman:
Yes, we should. Frank is under the weather but maybe next week Wednesday 14:00 to 15:00?

ruspl-afed:
next Wednesday looks good for me, but I need to understand the time zone of your proposed 14:00 to 15:00

wimjongman:
Yep. Sorry. I will send an invite to you and Frank.

PR #21, opened 2022-12-31
Fix the build for main branch #10

ruspl-afed:
Introduce org.eclipse.mylyn.parent and remove parent duplications

Signed-off-by: Alexander Fedorov <alexander.fedorov@arsysop.ru>


BeckerFrank (approved):
LGTM

Issue #22, opened 2023-01-01
target platform strategy


BeckerFrank:
This Issue was created in the now archived Repository org.eclipse.mylyn.commons by me
(see https://github.com/eclipse-mylyn/org.eclipse.mylyn.commons/issues/14)

ruspl-afed:
This one is done, target aligned with 2022-12 and has no circular dependencies

wimjongman:
Well done! Thank you.

PR #29, opened 2023-01-02
target platform strategy #22

ruspl-afed:
Remove target definitions from components

Signed-off-by: Alexander Fedorov <alexander.fedorov@arsysop.ru>


BeckerFrank (approved):
LGTM

PR #32, opened 2023-01-02
Add setup for .github repository

BeckerFrank:


org.eclipse.mylyn.releng/oomph/Mylyn.setup
```diff
@@ -424,6 +424,82 @@
         label="main"/>
     <description>The new Setup for Mylyn hosted on GitHub</description>
   </project>
+  <project name="github.admin"
+      label="Mylyn GitHub Organisation Administration">
+    <annotation
+        source="http://www.eclipse.org/oomph/setup/BrandingInfo">
```
wimjongman, line 430: use https instead of http? Here and elsewhere in the source.

BeckerFrank: The source of a annotation must be http:// and not https://

wimjongman (approved):
LGTM, but the proof is in the eating.

PR #40, opened 2023-01-20
Local jenkins

BeckerFrank:
Now I have a running local environment for Jenkins and Bugzilla Instances on Ubuntu.

You can use one of the following 3 setups
- swarm (3 VM's)
  this is a minimal Docker Swarm setup with one master and one worker node. The other vm only holds the rootca for the local setup
- container (2 VM's)
  Did not use Docker swarm, so we need no Swarm Worker Node 
- single (1 VM)
  store the rootca Service and files on the same VM as the docker

Can maybe someone not running on a Apple Silicon test if the setup works on Windows / Linux?


wimjongman (approved):
I propose to merge this and work on issues separately. @ruspl-afed WDYT Alexander?

ruspl-afed (approved):
+1
Licensing check failure is not related with this change

Issue #17, opened 2022-12-30
Drop Trac connector from mylyn.tasks


BeckerFrank:
This Issue was created in the now archived Repository org.eclipse.mylyn.tasks by Maarten Meijer
(see  https://github.com/eclipse-mylyn/org.eclipse.mylyn.tasks/issues/7)

Trac connector is included as second connector in source to show general application of mylyn.tasks framework.
To simplify maintenance drop the TRAC connector or spin out in separate mylyn.archive repository

BeckerFrank:
We did not remove the trac connector.
We now have the connector as a separate installable unit and refactored the source so that now the connectors a bundled in on folder and every connector is a subfolder.

PR #41, opened 2023-01-21
issue #17: Drop Trac Connector

BeckerFrank:
For now we do not delete the trac projects we only not import them into the workspace.

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.discovery.tests/plugin.xml
+++ b/mylyn.commons/org.eclipse.mylyn.discovery.tests/plugin.xml
@@ -30,65 +30,6 @@
             provider="UnitTesting"
             siteUrl="http://foo.nodomain/site">
       </connectorDescriptor>
-      <connectorDescriptor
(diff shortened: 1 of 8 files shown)


wimjongman (approved):

ruspl-afed (approved):
Well, I know people who still use trac.
But let's say they will have a starting point in a case of an irresistible desire to restore trac support

BeckerFrank:
> Well, I know people who still use trac.
> But let's say they will have a starting point in a case of an irresistible desire to restore trac support

Should we instead of the remove make Trac a extra installable feature?

wimjongman:
Task integrations are an important attractor. If people still use it, why drop it?

wimjongman:
> > Well, I know people who still use trac.
> > But let's say they will have a starting point in a case of an irresistible desire to restore trac support
> 
> Should we instead of the remove make Trac a extra installable feature?

Yes. Good idea. We can do the same for other integrations. Like Github.

BeckerFrank:
We continue this with pull request #42

Issue #53, opened 2023-02-14
launch files expect JavSE-1.6


gnl42:
The *.launch files wants JavaSe-1.6 and (removed parameters `-XX:PermSize=128m -XX:MaxPermSize=256m`

JavaSE-11 is the minimum needed for the current Eclipse/Mylyn combination to run/debug the code.

akurtakov:
Please note that 4.28 (2023-06) Eclipse Platform will require Java 17 so I would vote for Mylyn setting such a minimum too.

PR #54, opened 2023-02-14
launch files expect JavSE-1.6 #53

gnl42:
Short term fix until the base level has been decided on


wimjongman (approved):
Brilliant. Thanks, George.

PR #58, opened 2023-02-21
only the real changes the refactoring would be an other change

BeckerFrank:
First part of making Trac as an installable feature (see
[Trac Connector should be a installable feature](https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/17)


wimjongman (approved):
Looks good to me, Frank.

PR #59, opened 2023-02-22
refactor Task Connector locations

BeckerFrank:
Most files are only moved to new location, only some have changes!


ruspl-afed (approved):
To be honest I'm not a fan of deep hierarchical structures, but perhaps you have some clear picture in mind regarding where do we go in terms of repository layout.

wimjongman (approved):
Looks good to me. Let's go.

mjmeijer:
Where has the bugzilla connector gone? Trac is now optional, but Bugzilla has disappeared

BeckerFrank:
No Bugzilla is installed per Default and so we do not have an option but it is available.
Why do you think it is gone?
What url did you use?

mjmeijer:
Trying from the staging URL but install fails, will post details tomorrow

BeckerFrank:
Sorry wrong Information Here we have the Bugzilla Connector.
image
and same for https://download.eclipse.org/mylyn/stage/latest/

Issue #57, opened 2023-02-20
Hudson Connector - AuthenticationException: Forbidden


gnl42:
Created by Sebastian List: https://bugs.eclipse.org/bugs/show_bug.cgi?id=412596

=================================================================================
`Hi,
I experiencing authentication errors when trying to connect to a Hudson v3.0.0 server. I cannot refresh Build Plans, nor can I validate my settings. The Stacktrace reads as follows:

org.eclipse.core.runtime.CoreException: Unexpected error: Forbidden
	at org.eclipse.mylyn.internal.hudson.core.HudsonCorePlugin.toCoreException(HudsonCorePlugin.java:61)
	at org.eclipse.mylyn.internal.hudson.core.HudsonServerBehaviour.validate(HudsonServerBehaviour.java:908)
	at org.eclipse.mylyn.builds.internal.core.BuildServer.validate(BuildServer.java:369)
`

Sam Davis:
`hanks for the detailed bug report! It looks like the problem is that your server is not returning a 404 for /j_acegi_security_check - could you confirm what response you get when doing a GET to that URL? It's only when getting a 404 there that the connector tries /j_spring_security_check. Since we no longer support versions prior to 3.0.1 we should probably try the latter first and fall back to the former if it fails: https://git.eclipse.org/r/#/c/21655/. This appears to still work for Hudson 2.2.1.`

Sam Davis:
`Thanks. It sounds like providing the auth header might be a better solution, but it will be a while before we have time to look into that. It could help to speed this up if you submitted a code review. :)`

=================================================================================

I'm having the same problem with my Jenkins Docker instance, v2.391

Disable the 'CSRF Protecttion' "solution" is no longer an option.

wimjongman:
Thanks for pulling this in, George.

BeckerFrank:
Yes we need to change the authenticate to support CSRF Protection. This is a blocker for reactivate the HudsonTests

Issue #55, opened 2023-02-18
Hudson connector needs JAXB-API with Java 11+


gnl42:
Finally did some more exploring of the Hudson connector:

`
!ENTRY org.eclipse.mylyn.builds.ui 4 0 2023-02-18 01:36:33.085
!MESSAGE Server validation failed: Unexpected error: Implementation of JAXB-API has not been found on module path or classpath.
!STACK 1
org.eclipse.core.runtime.CoreException: Unexpected error: Implementation of JAXB-API has not been found on module path or classpath.
	at org.eclipse.mylyn.internal.hudson.core.HudsonCorePlugin.toCoreException(HudsonCorePlugin.java:62)
	at org.eclipse.mylyn.internal.hudson.core.HudsonServerBehaviour.refreshConfiguration(HudsonServerBehaviour.java:798)
	at org.eclipse.mylyn.builds.internal.core.operations.RefreshConfigurationOperation$1.run(RefreshConfigurationOperation.java:76)
	at org.eclipse.mylyn.builds.internal.core.operations.RefreshConfigurationOperation$1.run(RefreshConfigurationOperation.java:1)
	at org.eclipse.mylyn.builds.internal.core.util.BuildRunner.run(BuildRunner.java:73)
	at org.eclipse.mylyn.builds.internal.core.operations.RefreshConfigurationOperation.doRefresh(RefreshConfigurationOperation.java:73)
	at org.eclipse.mylyn.builds.ui.spi.BuildServerPart$Validator.run(BuildServerPart.java:140)
	at org.eclipse.mylyn.commons.repositories.ui.RepositoryLocationPart$2.run(RepositoryLocationPart.java:559)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:122)
Caused by: org.eclipse.mylyn.internal.hudson.core.client.HudsonException: Implementation of JAXB-API has not been found on module path or classpath.
	at org.eclipse.mylyn.internal.hudson.core.client.HudsonOperation.run(HudsonOperation.java:171)
	at org.eclipse.mylyn.internal.hudson.core.client.RestfulHudsonClient.getJobsFromFolder(RestfulHudsonClient.java:340)
	at org.eclipse.mylyn.internal.hudson.core.client.RestfulHudsonClient.getJobs(RestfulHudsonClient.java:267)
[...]

PR #60, opened 2023-02-23
Authorization/JAXB-API for Jenkins #55, #57

gnl42:
Added "crumbissuer" authentication. Do we still need the old authentication code?

Fixed missing JAXB-API complaint 
(https://www.vogella.com/blog/eclipse-rcp-java-11-jaxb/)


mylyn.builds/org.eclipse.mylyn.hudson.core/src/org/eclipse/mylyn/internal/hudson/core/client/HudsonOperation.java
```diff
@@ -70,11 +74,58 @@ protected String baseUrl() {
...
+								AuthenticationType.REPOSITORY));
+			}
+		} finally {
+			HttpUtil.release(request, response, monitor);
+		}
+
+	}
+
+	@Deprecated
+	private void legacyAuthentication(IOperationMonitor monitor, UserCredentials credentials)
```
wimjongman, line 127: You can delete this.

mylyn.builds/org.eclipse.mylyn.hudson.core/src/org/eclipse/mylyn/internal/hudson/core/client/HudsonOperation.java
```diff
@@ -109,14 +160,15 @@ protected void authenticate(IOperationMonitor monitor) throws IOException {
...
+			updateCrumb(monitor);
 		} finally {
 			HttpUtil.release(request, response, monitor);
 		}
-
-		updateCrumb(monitor);
 	}
 
-	public HttpResponse executeAuthenticationRequest(IOperationMonitor monitor, UserCredentials credentials,
+	@Deprecated
```
wimjongman, line 170: Please drop.

mylyn.builds/org.eclipse.mylyn.hudson.core/src/org/eclipse/mylyn/internal/hudson/core/client/HudsonOperation.java
```diff
@@ -70,11 +74,58 @@ protected String baseUrl() {
...
+						new AuthenticationRequest<AuthenticationType<UserCredentials>>(getClient().getLocation(),
+								AuthenticationType.REPOSITORY));
+			}
+		} finally {
+			HttpUtil.release(request, response, monitor);
+		}
+
+	}
+
+	@Deprecated
```
wimjongman, line 126: You can drop it.

mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
```diff
@@ -4,14 +4,21 @@ Bundle-Name: %Bundle-Name
...
-Require-Bundle: jakarta.xml.bind;bundle-version="2.3.3",
- org.eclipse.core.runtime;bundle-version="0.0.0",
+Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
  org.eclipse.emf.ecore;bundle-version="0.0.0",
  org.eclipse.mylyn.builds.core;bundle-version="3.26.0",
  org.eclipse.mylyn.commons.core;bundle-version="3.26.0",
  org.eclipse.mylyn.commons.repositories.core;bundle-version="3.26.0",
  org.eclipse.mylyn.commons.repositories.http.core;bundle-version="3.26.0",
- org.eclipse.mylyn.hudson.core;bundle-version="3.26.0"
+ org.eclipse.mylyn.hudson.core;bundle-version="3.26.0",
```
wimjongman, line 13: Can't you use import package instead of require bundle? Also, are all of these required?

mylyn.builds/org.eclipse.mylyn.hudson.core/src/org/eclipse/mylyn/internal/hudson/core/client/HudsonOperation.java
```diff
@@ -254,7 +327,7 @@ private boolean hasValidatAuthenticationState() {
 		List<Cookie> cookies = new ArrayList<Cookie>(getClient().getHttpClient().getCookieStore().getCookies());
 		if (cookies != null) {
 			for (Cookie cookie : cookies) {
-				if ("JSESSIONID".equals(cookie.getName())) { //$NON-NLS-1$
+				if ("JSESSIONID".equals(cookie.getName()) || cookie.getName().startsWith("JSESSIONID")) { //$NON-NLS-2$
```
wimjongman, line 330: Put in JSESSIONID in a constant?

wimjongman (changes requested):
Thanks, George! 

This looks good. I found a couple of nits.

mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
```diff
@@ -4,14 +4,18 @@ Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.mylyn.hudson.core;singleton:=true
 Bundle-Version: 3.26.0.qualifier
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: jakarta.xml.bind;bundle-version="2.3.3",
- org.eclipse.core.runtime;bundle-version="0.0.0",
+Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
```
ruspl-afed, line 7: Please keep alphabetic order for dependencies

org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -71,6 +71,22 @@
...
+				    <artifactId>jaxb-impl</artifactId>
+				    <version>2.3.3</version>
+				</dependency>
+				<dependency>
+				    <groupId>jakarta.xml.bind</groupId>
+				    <artifactId>jakarta.xml.bind-api</artifactId>
+				    <version>2.3.3</version>
+				</dependency>
+				<dependency>
+				    <groupId>org.glassfish.hk2</groupId>
```
ruspl-afed, line 85: Why do we need this one?

org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -71,6 +71,22 @@
 		</location>
 		<location includeDependencyDepth="infinite" includeDependencyScopes="compile,provided,runtime" includeSource="true" missingManifest="generate" type="Maven">
 			<dependencies>
+				<dependency>
+				    <groupId>com.sun.xml.bind</groupId>
+				    <artifactId>jaxb-impl</artifactId>
+				    <version>2.3.3</version>
+				</dependency>
+				<dependency>
+				    <groupId>jakarta.xml.bind</groupId>
```
ruspl-afed, line 80: we have this in the target already from `https://download.eclipse.org/tools/orbit/downloads/drops/R20221123021534`

org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -71,6 +71,22 @@
 		</location>
 		<location includeDependencyDepth="infinite" includeDependencyScopes="compile,provided,runtime" includeSource="true" missingManifest="generate" type="Maven">
 			<dependencies>
+				<dependency>
+				    <groupId>com.sun.xml.bind</groupId>
```
ruspl-afed, line 75: Would be better to use one from `https://download.eclipse.org/tools/orbit/downloads/drops/R20221123021534`

ruspl-afed (changes requested):

gnl42:
Hi Wim, 
  I made the changes and had to rework the code a bit to fix some issues I came across after some more testing.

The build framework is weird, I don't understand what the intent was in its design.
CommonHttpClient has an "authenticated" flag that is good for a session, yet has a ThreadLocal variable "context" that is used to store the token. Side effect of this is that each thread had to authenticate rather than just doing it once for the session.

One issue I'm unable to solve is how use the JSESSION id to authenticate when running a build using ones password.
It works fine when using a API token, but according to all the documentation I've come across it should be doable in both cases???

I've been testing it against my local Jenkins instance, and against https://ci.eclipse.org/mylyn/job/github-mylyn/.
If someone who is able to login to ci.eclipse.org could test the code I would feel more comfortable.

Also, I haven't seen any sign that the crumb is needed. Could be my jenkins hasn't been set up fully.

org.eclipse.mylyn-target/mylyn-e4.26.target
gnl42: I have no idea how it's all wired together. Maybe https://www.vogella.com/blog/eclipse-rcp-java-11-jaxb/ will make more sense to you?

ruspl-afed: Yes, true, seems to be required to activate plain java services expected by JAXB impl with OSGi. We can get this bundle from https://download.eclipse.org/tools/orbit/downloads/drops/R20221123021534/ as well.

BeckerFrank:
> Also, I haven't seen any sign that the crumb is needed. Could be my jenkins hasn't been set up fully.

I have finally remember the password so that I could reactivate the Hudson and Jenkins test instances of the http://mylyn.org domain. Maybe that helps.

org.eclipse.mylyn-target/mylyn-e4.26.target
gnl42: Hmm, when I moved it to orbit, the build breaks with:
`Error:  Cannot resolve project dependencies:
Error:    Software being installed: org.eclipse.mylyn.commons.feature.group 3.26.0.qualifier
Error:    Missing requirement: org.eclipse.mylyn.commons.ui 3.26.0.qualifier requires 'osgi.bundle; org.apache.commons.lang [2.6.0,3.0.0)' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.mylyn.commons.feature.group 3.26.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.mylyn.commons.ui 0.0.0`

gnl42: Never mind. It's only the source that's available from orbit. Sticking to the maven version for now

gnl42: Orbit searches are weird. I did find the right one, but it has the same problem.
I also tried the 2.3.0 version, but it has similar, but different, problems

org.eclipse.mylyn-target/mylyn-e4.26.target
gnl42: Hmm, when I moved it to orbit, the build breaks with:
Error:  Cannot resolve project dependencies: Error:    Software being installed: org.eclipse.mylyn.commons.feature.group 3.26.0.qualifier Error:    Missing requirement: org.eclipse.mylyn.commons.ui 3.26.0.qualifier requires 'osgi.bundle; org.apache.commons.lang [2.6.0,3.0.0)' but it could not be found Error:    Cannot satisfy dependency: org.eclipse.mylyn.commons.feature.group 3.26.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.mylyn.commons.ui 0.0.0

I also tried the 2.3.0 version, but it has similar, but different, problems

gnl42: Tried to change it to commons.lang3 but it's not available from orbit(?)

mylyn.builds/org.eclipse.mylyn.hudson.core/src/org/eclipse/mylyn/internal/hudson/core/client/HudsonOperation.java
gnl42: Alas, the Hudson/Jenkins servers used for the unit tests doesn't support the new way, so I put them back

gnl42:
> > Also, I haven't seen any sign that the crumb is needed. Could be my jenkins hasn't been set up fully.
> 
> I have finally remember the password so that I could reactivate the Hudson and Jenkins test instances of the http://mylyn.org domain. Maybe that helps.

If that is the same instances that the unit tests are using (e.g.: https://mylyn.org/hudson-3.3.3, http://mylyn.org/jenkins-2.32.3/) it doesn't help.

Neither of them accept the ''crumbIssuer/api/json" request

BeckerFrank:
Maybe the versions are to old to support that or we have never configure this.

gnl42:
@BeckerFrank 
> Maybe the versions are to old to support that or we have never configure this.

Looks like 2.32.3 was released back in 2016 so I would say it's too old.
It also hasn't been configured to use the crumb security (CSRF not enabled)

wimjongman:
> I would suggest merging the code and deal with any issues that crop up with a new ticket

Hey George. Cool! Do you want to take a look at the review comments before merging?

org.eclipse.mylyn-target/mylyn-e4.26.target
ruspl-afed: OMG. Let's postpone this improvement, I'd rather have a closer look on our .target later.

ruspl-afed (approved):
I've removed my comments regarding issues in .target, I need to have closer look later.

gnl42:
> > I would suggest merging the code and deal with any issues that crop up with a new ticket
> 
> Hey George. Cool! Do you want to take a look at the review comments before merging?

I think I've addressed the comments to the best of my ability. When I tried to address the comments about the .target I ran into  unresolved references that I have no idea how to fix

gnl42:
Don't merge yet. Started looking into the session Id problem and found some very not nice behaviour

gnl42:
Solved the "sessionId" issue.

user/pw can run build now. Caveat is that if Jenkins is using StrictCrumbIssuer with a timeout set for the crumb one will be unable to run builds once the crumb expires. I can't see any way to have a new crumb created in this situation with how the build framework is designed.
Does not happen with a token, the recommended way, so I'm ok with this behaviour.

wimjongman:
George, is the token used in the password field?

gnl42:
Hi Wim,
   yes.

It's getting complicated. Depending on the system, authentication can be by:

  * Token
  * username/pw
  * username/token

George

wimjongman:
Yes, it is. To be clear by 'system' do you mean systems like Jenkins or different Jenkins versions? Is the UI still using "Hudson"?

wimjongman (approved):

gnl42:
Yes,

  * Token: Bamboo 8, GitHub, Jira 8
  * username/pw: Bamboo 8, Jenkins, Jira 8
  * username/token: Jira 10, Jenkins (modern)

The connector is advertised as 'Hudson (supports Jenkins)' but I don't 
see anything in the UI itself

BeckerFrank:
> @BeckerFrank
> 
> > Maybe the versions are to old to support that or we have never configure this.
> 
> Looks like 2.32.3 was released back in 2016 so I would say it's too old. It also hasn't been configured to use the crumb security (CSRF not enabled)
Yes we should use a newer Jenkins version in our new test infrastructure (see Multipass setup supports >= 2_303_3)

gnl42:
Could someone merge the PR?

PR #61, opened 2023-02-25
fix: Startup Message General problem with descriptor entry

BeckerFrank:
for "!MESSAGE bundle org.eclipse.mylyn.commons.identity.core:3.26.0.qualifier (608) General problem with descriptor entry '/OSGI-INF/component.xml' "
As a Result we do not see the pictures in UserAttributeEditor


ruspl-afed (approved):
Later we could setup project to generate component descriptor using annotation

PR #68, opened 2023-03-18
Mylyn 3.26 support for 2022-06 or later  

BeckerFrank:
Make sure that Mylyn 3.26.0 runs on Eclipse 2022-06 or later.


org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -2,128 +2,131 @@
 <?pde version="3.8"?>
 <target name="mylyn-e4.26">
 	<locations>
-		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/cbi/updates/license/"/>
-			<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
+		<location includeAllPlatforms="false" includeConfigurePhase="false"
```
ruspl-afed, line 5: Do you think we can reduce changes for .target @BeckerFrank ?

BeckerFrank: Sorry I found that the unit tests for Hudson fail because of missing components at runtime. So actual I try to fix this and I did not active change this line but look what is changed for my next commit.

wimjongman: Frank, what about org.eclipse.egit.github.core. It is not in the target platform. Did you perform some maven magic to make it run?

BeckerFrank:
Actual I can not install the Jenkins Connector because of a missing dependency. This also happens with the snapshot build from the main branch.

Think we should first need to resolve this issue before we can continue here.

org.eclipse.mylyn-target/mylyn-e4.26.target
BeckerFrank: > Frank, what about org.eclipse.egit.github.core. It is not in the target platform. Did you perform some maven magic to make it run?

No no magic it is in the type maven part. Here we add the repository https://repo.eclipse.org/content/repositories/egit-releases/ and use the maven coordinate 
				<dependency>
					<groupId>org.eclipse.mylyn.github</groupId>
					<artifactId>org.eclipse.egit.github.core</artifactId>
					<version>6.1.0.202203080745-r</version>
					<type>jar</type>
				</dependency>

ruspl-afed: > So actual I try to fix this and I did not active change this line but look what is changed for my next commit.

Perhaps, there are formatting changes there, It would be great if you can revert them.

BeckerFrank:
@wimjongman, @ruspl-afed: Think we can now continue with the review.

wimjongman (approved):
LGTM! Thanks.

PR #69, opened 2023-03-20
Hudson connector needs JAXB-API with Java 11+ #55

ruspl-afed:
Switch from `Require-Bundle` to `Import-Package`for javax.xml 
Switch from includes/requires for feature to providing in repository

### DIFF
--- a/mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
+++ b/mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
@@ -23,7 +23,6 @@ Export-Package: org.eclipse.mylyn.hudson.core;x-friends:="org.eclipse.mylyn.huds
 Bundle-Activator: org.eclipse.mylyn.internal.hudson.core.HudsonCorePlugin
 Import-Package: com.google.gson;version="[2.7.0,2.9.2)",
  com.sun.xml.bind;version="2.3.3",
- jakarta.activation;version="2.0.0",
  javax.xml.bind;version="2.3.3",
  javax.xml.bind.annotation;version="2.3.3",
  org.apache.http;version="[4.4.4,4.5.0)",
--- a/mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
+++ b/mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
@@ -4,9 +4,7 @@ Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.mylyn.hudson.core;singleton:=true
 Bundle-Version: 3.26.0.qualifier
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: jakarta.xml.bind;bundle-version="2.3.3",
- jakarta.xml.bind-api;bundle-version="2.3.3",
- org.apache.commons.io;bundle-version="2.8.0",
+Require-Bundle: org.apache.commons.io;bundle-version="2.8.0",
  org.apache.httpcomponents.httpcore;bundle-version="4.4.15",
  org.eclipse.core.runtime;bundle-version="0.0.0",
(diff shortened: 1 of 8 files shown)


mylyn.builds/org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF
```diff
@@ -26,6 +24,8 @@ Bundle-Activator: org.eclipse.mylyn.internal.hudson.core.HudsonCorePlugin
 Import-Package: com.google.gson;version="[2.7.0,2.9.2)",
  com.sun.xml.bind;version="2.3.3",
  jakarta.activation;version="2.0.0",
```
BeckerFrank, line 26: I switch to the 55_xlm branch and did a local build. When I use the local repository I get the following error.

```
Cannot complete the install because one or more required items could not be found.
  Software being installed: Mylyn Builds Connector: Hudson/Jenkins 3.26.0.N20230320-1407 (org.eclipse.mylyn.hudson.feature.group 3.26.0.N20230320-1407)
  Missing requirement: Mylyn Hudson Connector 3.26.0.N20230320-1407 (org.eclipse.mylyn.hudson.core 3.26.0.N20230320-1407) requires 'java.package; jakarta.activation 2.0.0' but it could not be found
  Cannot satisfy dependency:
    From: Mylyn Builds Connector: Hudson/Jenkins 3.26.0.N20230320-1407 (org.eclipse.mylyn.hudson.feature.group 3.26.0.N20230320-1407)
    To: org.eclipse.equinox.p2.iu; org.eclipse.mylyn.hudson.core [3.26.0.N20230320-1407,3.26.0.N20230320-1407]
```

I need to delete the line 
 jakarta.activation;version="2.0.0",
from /org.eclipse.mylyn.hudson.core/META-INF/MANIFEST.MF to resolve this error.

gnl42:
This is very confusing. Can someone explain why unresolved dependencies appear/disappear seemingly randomly?
Is there a document somewhere that explains how OSGI resolution  works?

ruspl-afed:
I think we have here both OSGi (for runtime/tests) and p2 (for install time) dark magic in play.
Unfortunately the volume of reading regarding the subj could exceed reasonable timeframe, So I'd better try to explain what has been done here and why:
1)  switching from `Require-Bundle` to `Import-Package` gives more freedom to resolve required package that could be supplied by different available bundles
2) shifting "library" bundle (like `javax.*`) from feature.xml to p2 repo allows us to reuse compatible bundle from the pool of already installed bundles, this is important when installing/updating for different installation basis.

Not sure why `jakarta.activation;version="2.0.0"` was required - if we still need it, let's add a test that will fail because of its absence.

ruspl-afed:
> Is it possible to have a connector specific target file as well as the global one? So we could isolate certain libraries from "infecting" the rest of the world

It is quite tricky, but possible. We can isolate connector on build time, but how will it installed and then work together with others on runtime in case of dependency conflict?

gnl42:
I experimented a bit and was able to remove all explicit references to jakarta.activation

gnl42:
I would like to propose #71  as an alternative solution

ruspl-afed:
> I would like to propose https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/71 as an alternative solution

Since it requires more work to stabilize Lucene compatibility can we merge this one first and then continue with #71 ?

gnl42:
Go ahead. I'm cancelling #71 and will think a bit more about it.

Upgrading jaxb to 4.0 showed a lot of promise (jakarta instead of javax, less configuration), but in the end when I tried to install the connector it was unable to locate the needed libraries. It ran/debugged fine so frustrating.

Switching lucene to 8.4.1 fixes the failure to start the test but tests fail with the 'cannot change field "creation_date" from index options=DOCS to inconsistent index options=DOCS_AND_FREQS_AND_POSITIONS' complaint

ruspl-afed:
> Go ahead.

Thanks @gnl42
Could you please approve this PR @BeckerFrank @wimjongman ?

wimjongman (approved):
LG!

Issue #70, opened 2023-03-21
org.eclipse.mylyn.tasks.activity.tests fails


gnl42:
mvn clean verify
`[ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:3.0.3:test (default-test) on project org.eclipse.mylyn.tasks.activity.tests: An unexpected error occurred while launching the test runtime (process returned error code -1 (HRESULT Code 0xFFFFFFFF`

```
!ENTRY org.eclipse.mylyn.tasks.index.core 4 0 2023-03-20 21:27:33.844
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.mylyn.tasks.index.core [97]
  Bundle was not resolved because of a uses constraint violation.
  org.apache.felix.resolver.reason.ReasonException: Uses constraint violation. Unable to resolve resource org.eclipse.mylyn.tasks.index.core [osgi.identity; osgi.identity="org.eclipse.mylyn.tasks.index.core"; type="osgi.bundle"; version:Version="3.26.0.N20221223-1911"] because it is exposed to package 'org.apache.lucene.analysis' from resources org.apache.lucene.analyzers-common [osgi.identity; osgi.identity="org.apache.lucene.analyzers-common"; type="osgi.bundle"; version:Version="8.4.1.v20221112-0806"] and org.apache.lucene.core [osgi.identity; osgi.identity="org.apache.lucene.core"; type="osgi.bundle"; version:Version="6.1.0.v20170814-1820"] via two dependency chains.

Chain 1:
  org.eclipse.mylyn.tasks.index.core [osgi.identity; osgi.identity="org.eclipse.mylyn.tasks.index.core"; type="osgi.bundle"; version:Version="3.26.0.N20221223-1911"]
    require: (&(osgi.wiring.bundle=org.apache.lucene.analyzers-common)(bundle-version>=6.1.0))
     |
    provide: osgi.wiring.bundle: org.apache.lucene.analyzers-common
  org.apache.lucene.analyzers-common [osgi.identity; osgi.identity="org.apache.lucene.analyzers-common"; type="osgi.bundle"; version:Version="8.4.1.v20221112-0806"]

Chain 2:
[...]

PR #71, opened 2023-03-22
Update jaxb, lucene libraries #55, #70

gnl42:
Clean up jaxb-impl usage by Hudson connector, fix version mismatch with lucene


gnl42:
Also seems to make the target file less likely to introduce unexpected package resolution errors, don't know why.

ruspl-afed:
Not sure what kind of problem we are solving here. 

1) Why do we need to update JAXB to 4.x right now?

2) Why do we need to update Lucene here? Is it related to Hudson connector?

gnl42:
Needs more thinking about

PR #73, opened 2023-03-29
remove the old Trac Workingset

BeckerFrank:


BeckerFrank:
@wimjongman , @ruspl-afed: Any idea why we have a license checking problem when only changing [Mylyn.setup](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/73#diff-990c1078e104ec92925ec088de16bfb0fd0a93ff3313df71d28795d5ccca33e9)

ruspl-afed:
We suddenly have got troubles with
`
[INFO] maven/mavencentral/org.glassfish.hk2/osgi-resource-locator/2.4.0
[INFO] p2/orbit/p2.eclipse.plugin/org.objectweb.asm/9.4.0.v20221107-1714
`

Perhaps something has been changed recently for Dash Tool database. Could you please comment on this @waynebeaton ?

waynebeaton:
We had to tune the way that content is identified as _project content_ and it has some unintended (and some intended) consequences. 

`maven/mavencentral/org.glassfish.hk2/osgi-resource-locator/2.4.0` being rejected is an unintended consequence. But... it's actually helpful that it was identified. This version of the content is from before the GlassFish project was moved over to the Eclipse Foundation, so the license on it was previously reported incorrectly. I've used this opportunity to update the license information.

I should point out that `org.glassfish.hk2:osgi-resource-locator:2.4.0` is at least six years old at this point; I recommend updating to more modern version.

`p2/orbit/p2.eclipse.plugin/org.objectweb.asm/9.4.0.v20221107-1714` is a little more messed up. In this case, the tool had previously identified this a content that comes from Orbit and approved it on that basis alone. Which sort of put us in a bit of a loop since it being in Orbit is enough to approve it to be in Orbit (which is just weird and wrong). We actually want this to be run through the full IP due diligence process (which I've done).

In any case, both should pass now (with some changes in the licenses and authority reported in the summary).

ruspl-afed (approved):
LGTM

wimjongman:
Thanks, Wayne.

Issue #78, opened 2023-04-04
GitHub Issues: NPE when opening issue as Anonymous


gnl42:
Create a GitHub Issue repository as Anonymous, try open an issue:

`ENTRY org.eclipse.mylyn.tasks.ui 4 0 2023-04-03 21:47:12.034
!MESSAGE Could not create editor via factory: org.eclipse.mylyn.internal.github.ui.issue.IssueTaskEditorPageFactory@20505460
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.mylyn.commons.net.AuthenticationCredentials.getUserName()" because "cred" is null
	at org.eclipse.mylyn.internal.github.ui.issue.IssueTaskEditorPage.checkCanSubmit(IssueTaskEditorPage.java:109)
	at org.eclipse.mylyn.internal.github.ui.issue.IssueTaskEditorPage.createParts(IssueTaskEditorPage.java:87)
	at org.eclipse.mylyn.tasks.ui.editors.AbstractTaskEditorPage.createFormContentInternal(AbstractTaskEditorPage.java:685)
	at org.eclipse.mylyn.tasks.ui.editors.AbstractTaskEditorPage.createFormContent(AbstractTaskEditorPage.java:630)
	at org.eclipse.ui.forms.editor.FormPage.lambda$0(FormPage.java:166)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
`

PR #79, opened 2023-04-04
Fix NPE associated with GitHub Isses when Anonymous #78

gnl42:
Don't try to use a null pointer


wimjongman (approved):
LGTM.

wimjongman:
I think George is a committer now. Welcome, George!

gnl42:
I think there is still some paper work to finish:


but thanks

Issue #66, opened 2023-03-17
Set java version to 17


gnl42:
Can we set the java level to 17 given than Eclipse has been shipping with 17 for the last couple of releases, and 17 will be the base level for Eclipse 2023-06 (https://www.eclipse.org/lists/platform-dev/msg03746.html)

akurtakov:
FYI it's already changed in I-builds https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/fb5049c9570a9d2357bff64faf0db7203fbd8244 and there are already bundles requiring Java 17. 
+1000 from me for this.

gnl42:
Can we at least start the process and move to 11 now?

BeckerFrank:
Maybe we should look if we really need to include the 3 open issues in the [3.26.0](https://github.com/eclipse-mylyn/org.eclipse.mylyn/milestone/1) release or if we can start the first new release.

gnl42:
It's a bit silly in that I need Java 11 to lunch Mylyn for debugging/testing yet the code is at 1.8

wimjongman:
George, please go ahead if you want to make a PR to jump to 17.

gnl42:
Hi Wim,
    sure, but if it requires any changes for the Jenkins builds I will 
need guidance

PR #80, opened 2023-04-04
Use JavaSE-17 #66

gnl42:
Switch to Java 17. Purely a brute force approach with checking.
Doesn't seem to introduce any new issues (current ones are still there for me).

Would be nice if Mylyn.setup could be updated to set the cleanup/save Eclipse preferences for Java 17, but I have no idea how to do that (yet).


pdulth:
Hi, there is also the 1.8 to 17 occurences in **/.settings/org.eclipse.jdt.core.prefs i think

wimjongman:
Java 17 support will be the first thing after 3.26.1

The reason is that Mylyn has been broken for a longer time. Moving to 17 would prevent everything pre-3.27 from consuming a working copy of Mylyn.

BeckerFrank:
@gnl42: can you please update the branch so that we no longer have conflicts.

gnl42:
It was easier to recreate the changes than trying to figure out what went wrong with the merge. 
Also a cleaner PR.

PR #81, opened 2023-04-08
70 org.eclipse.mylyn.tasks.activity.tests fails

BeckerFrank:
With this we only use org.apache.lucene version to be less then 8.0 so we do not migrate the data and can stay with the tested 6.1 version.

To get the build work I need to switch to apache-maven-3.8.6 (from apache-maven-latest) Maven 3.9 did not work and 3.9.1 is actual not available.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ pipeline {
 		label "centos-latest"
 	}
 	tools {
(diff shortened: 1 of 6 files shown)


BeckerFrank:
This is the fix for #70

wimjongman (approved):
LGTM! Thanks, Frank.

PR #87, opened 2023-04-13
Resource filters defined by Oomph setup should be pushed #86

### LINKED ISSUE #86: Resource filters defined by Oomph setup should be pushed to repository
Currently after using Oomph script we have several outgoing changes caused by resource filter applied by Oomph setup.

Let's push corresponding changes to the repository and remove them from Mylyn.setup to have "up-to-date" source state after setup.

ruspl-afed:
Move filters from Mylyn.setup to .project manifests

Fixes #86


wimjongman (approved):
LGTM. Thanks. Does this solve George's problem?

ruspl-afed:
> LGTM. Thanks. Does this solve George's problem?

I hope so. And it solves mine.

Issue #88, opened 2023-04-13
Employ dependabot for org.eclipse.mylyn repository


ruspl-afed:
Release Engineering work may yet the whole chunk of resources available for a project and it should be automated as much as possible.

Let's employ [dependabot](https://github.com/dependabot) to help us with dependencies management. It already does great job for dependencies from `pom.xml` . 
And there is a [chance](https://github.com/dependabot/dependabot-core/pull/6913) (_**please don't forget to add your supporting comment there**_) that it will be able to manage `.target` manifests as well.

ruspl-afed:
Done with #89

ruspl-afed:
@waynebeaton please have a look at https://github.com/eclipse/dash-licenses/issues/225, all the PRs created by dependabot are blocked with license check error

ruspl-afed:
@waynebeaton we still need your assistance to understand/fix dash-licenses tool failures

ruspl-afed:
Perhaps I've found what is wrong, the configured project id is `technology.mylyn` while it should be `tools.mylyn`
I wonder why does it matter only for PRs created by dependabot and has no matter for others.

ruspl-afed:
Fixed with #98

PR #98, opened 2023-04-20
Employ dependabot for org.eclipse.mylyn repository #88

ruspl-afed:
Correct project id


akurtakov (approved):
Definitely better to have the correct projectId.

Issue #103, opened 2023-05-01
Manage External Dependencies


merks:
I'm concerned about how external dependencies are managed and in particular about any of the direct-from-maven dependencies that are **generated**.   It's hard to tell which ones in the list are generated versus which ones are actually already well-formed OSGi artifacts:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/f8a6aa79cac4838c788101fd10f470971ba6bdce/org.eclipse.mylyn-target/mylyn-e4.26.target#L69-L141

I really don't see how we can properly coordinate versions of libraries  in SimRel when projects independently of each other  generate wrappers for non OSGi-libraries.  What's generated depends on the BND recipe:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/f8a6aa79cac4838c788101fd10f470971ba6bdce/org.eclipse.mylyn-target/mylyn-e4.26.target#L132-L141

Each project may and likely will use a different recipe which appears likely to result in many different artifacts that use the same key such that they appear to be the same artifact from a p2 perspective but are actually different.   That's not just a problem for SimRel but a problem in general when installing such content into any installation.

Here's an example:

https://github.com/eclipse/tm4e/blob/ac34c5f440f95e1046ecb69b1c3c5c8514cb4a86/target-platform/tm4e-target.target#L44-L52

Making all the package imports optional in the recipe is convenient but is highly likely to be wrong:
```
Import-Package:        *;resolution:=optional
```

In any case, these artifacts are not even being PGP signed as far as I can tell:

https://download.eclipse.org/oomph/archive/reports-extra/mylyn/download.eclipse.org/mylyn/stage/latest/index.html

![image](https://user-images.githubusercontent.com/208716/235417047-2222ba54-42a5-48a5-bb15-4347a0d2d206.png)

It's also strange to see EGit's GitHub integration come from a Maven repository rather than more directly from EGit's downloads:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/f8a6aa79cac4838c788101fd10f470971ba6bdce/org.eclipse.mylyn-target/mylyn-e4.26.target#L126-L130

https://download.eclipse.org/egit/github/

I feel that these generated direct-from-maven dependencies are opening up a new can of worms, and we've not even been able to fully manage the existing can of worms...

ruspl-afed:
[...]

PR #104, opened 2023-05-01
[#103] Manage External Dependencies

ruspl-afed:
Remove unnecessary maven egit repo

### DIFF
--- a/org.eclipse.mylyn-target/mylyn-e4.26.target
+++ b/org.eclipse.mylyn-target/mylyn-e4.26.target
@@ -66,6 +66,22 @@
 			<unit id="org.tigris.subversion.clientadapter.feature.feature.group" version="1.8.6"/>
 			<unit id="org.tigris.subversion.subclipse.feature.group" version="1.8.22"/>
 		</location>
+		<location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" label="EGit GitHub" missingManifest="error" type="Maven">
+			<dependencies>
+				<dependency>
+					<groupId>org.eclipse.mylyn.github</groupId>
+					<artifactId>org.eclipse.egit.github.core</artifactId>
+					<version>6.1.0.202203080745-r</version>
+					<type>jar</type>
+				</dependency>
+			</dependencies>
+			<repositories>
+				<repository>
+					<id>eclipse-egit</id>
+					<url>https://repo.eclipse.org/content/repositories/egit-releases/</url>
+				</repository>
+			</repositories>
+		</location>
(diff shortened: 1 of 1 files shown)


org.eclipse.mylyn-target/mylyn-e4.26.target
```diff
@@ -123,12 +123,6 @@
 					<type>jar</type>
 				</dependency>
 			</dependencies>
-			<repositories>
```
BeckerFrank, line 126: Where do wie then find the org.eclipse.egit.github.core plugin which is used by org.eclipse.mylyn.github.core?

ruspl-afed: My bad, fixed

BeckerFrank: Will post soon an update after I have tried that https://download.eclipse.org/egit/github/updates as p2 works.

ruspl-afed: Please reference the exact bundle as IU if you want to switch there

ruspl-afed: And then please merge this one @BeckerFrank if it works for you

BeckerFrank: > Will post soon an update after I have tried that https://download.eclipse.org/egit/github/updates as p2 works.

Sorry this did not work because the plugin org.eclipse.egit.github.core_6.1.0.202203080745-r.jar is not installable with  a separate feature. It is included in the feature org.eclipse.mylyn.github.feature_6.1.0.202203080745-r.jar

We need to request that org.eclipse.egit.github.core should be a standalone feature.

If we need to change the target file and remove the maven location I think we should try.
download the needed jar into a folder parallel to the target file and use Type Directory instead of Maven.
But for this case, I can't find a value for the path that works for a maven build and for the running eclipse installation.

wimjongman:
Hey let me know when to test. I was away for a conference but back again. Just tried to install gh connector but it fails trying to uninstall egit.

BeckerFrank:
@wimjongman , @ruspl-afed:
I think we can merge this!
I was the last who changed the file so maybe someone else can do the review and merge.

Actual I do a combined test for #104 and #105 on my local env to verify that the update site and installation are ok.

wimjongman (approved):
LGTM.

ruspl-afed:
I cannot add a review/approval since I was a creator.

BTW, how about to allow Project Leads merge without required approval? It may be required for some cases.

PR #106, opened 2023-05-02
[#103] Manage External Dependencies

### LINKED ISSUE #103: Manage External Dependencies

ruspl-afed:
Fixes #103

### DIFF
--- a/org.eclipse.mylyn-target/mylyn-e4.26.target
+++ b/org.eclipse.mylyn-target/mylyn-e4.26.target
@@ -31,7 +31,7 @@
 			<unit id="org.eclipse.cdt.sdk.feature.group" version="0.0.0"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20180206163158/repository"/>
+			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20180206163158/repository/"/>
 			<unit id="com.google.gerrit.common" version="0.0.0"/>
 			<unit id="com.google.gerrit.prettify" version="0.0.0"/>
 			<unit id="com.google.gerrit.reviewdb" version="0.0.0"/>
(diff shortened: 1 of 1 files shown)


ruspl-afed:
@merks I hope this is the last part of #103 , custom recipes were eliminated

wimjongman (approved):

PR #107, opened 2023-05-03
Get org.eclipse.egit.github.core from a p2 repository

merks:


merks:
@ruspl-afed 

It's not a big deal, but I think getting this bundle directly from EGit's p2 repository is better than relying on was is being published to maven central...

ruspl-afed (approved):
Yes, you are right. Thank you @merks

PR #108, opened 2023-05-03
Change build qualifier from `N` to `v`

ruspl-afed:
Perhaps we inherited `N` from nightly, but it is more common to see `v` as a build qualifier prefix
WDYT @merks ?

### DIFF
--- a/mylyn.releng/org.eclipse.mylyn.parent/pom.xml
+++ b/mylyn.releng/org.eclipse.mylyn.parent/pom.xml
@@ -54,7 +54,7 @@
     <sign.skip>true</sign.skip>
     <pack.skip>false</pack.skip>
     <promote.skip>true</promote.skip>
-    <dist.qualifier>'N'yyyyMMdd-HHmm</dist.qualifier>
+    <dist.qualifier>'v'yyyyMMdd-HHmm</dist.qualifier>
     <dist.repository>${project.name}</dist.repository>
     <dist.archive>${basedir}/target/drops</dist.archive>
     <dist.site></dist.site>


merks:
Yes v is more common. Be very careful and aware that the qualifier is compared lexicographically so I think even cases matters.   You don’t want to make a new version that’s less.

ruspl-afed:
> You don’t want to make a new version that’s less.

This is what I was secretly afraid of. Thank you for your warning @merks 
I would better postpone this change till 3.26.1 to not ruin what we have already.

merks:
Note that N < v even if one ignored case so I don’t think that would cause a problem.

ruspl-afed:
> Note that N < v even if one ignored case so I don’t think that would cause a problem.

Who knows what encapsulation hides from us ...

wimjongman (approved):
LGTM!

ruspl-afed:
We've found [this](https://download.eclipse.org/mylyn/drops/3.26.0/v20200731-0526/) location and decided with @wimjongman  to merge this one, otherwise comparison will work against us.

BeckerFrank:
> We've found [this](https://download.eclipse.org/mylyn/drops/3.26.0/v20200731-0526/) location and decided with @wimjongman to merge this one, otherwise comparison will work against us.

Maybe that was my mistake back in 2020.
If I remember correctly, I start building a 3.26 release and then we move to another 3.25.x service release.

merks:
It's water under the bridge...

BeckerFrank:
Please let us put the dist.qualifier into a profile and use this in the maven pipeline build.
We need a default that we can use for a local build and also set the dist.timestampProvider to empty.
Otherwise we can have locale create artifacts that conflicts with artifacts from a staging repository.

ruspl-afed:
@BeckerFrank I'm not sure I understood your suggestion well. Please create an issue or/and PR so we can discuss it there in details.

wimjongman:
> @BeckerFrank I'm not sure I understood your suggestion well. Please create an issue or/and PR so we can discuss it there in details.

I already created #113

PR #115, opened 2023-05-04
Add in a version of logging that is compatible

jonahgraham:
xmlrpc.server requires a logging < 1.2.0

This resolves the immediate issue in SimRel - https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/201693/comments/cf5c3a34_9e765dd6

I did a `mvn clean package` and then then added `$PWD/org.eclipse.mylyn-site/target/repository` to the simrel.aggr and validated locally.


ruspl-afed (approved):
Thank you so much @jonahgraham

Issue #121, opened 2023-05-06
Prepare releng for 3.26.1


ruspl-afed:
We need to prepare releng to the next release cycle:
* increment version in manifests
* update target platform
* and so on

BeckerFrank:
The old way was to change and run org.eclipse.mylyn.releng/bin/update-versions.sh.

ruspl-afed:
yes, I've seen it, but I was not pretty sure what will happen, since it needs some update and debugging first, and this is problematic to do on Windows.

I can recognize a solid process behind all these scripts. Unfortunately, this knowledge seems to be lost. 

Since there is no queue of volunteers dreaming of completing all the releng tasks on time, I need to do these task with my own poor understanding of the glorious legacy. But I do not insist, if someone could do it better, I would love to learn from this person.

Do you know if we introduce some defects when we do not use org.eclipse.mylyn.releng/bin/update-versions.sh @BeckerFrank ?

BeckerFrank:
When I tried to produce the 3.25.x service release I used [this](https://wiki.eclipse.org/Mylyn/Release_Howto) an a source of information.

ruspl-afed:
After migration to GitHub and merging repositories these instructions and scripts cannot be used "as is".  
I hope that @merks is going to provide us an update in the scope of #116

merks:
@ruspl-afed 

Sorry I have it mostly ready and I did a few changes to the JustJ publisher so it works better for Mylyn, but I keep getting tugged away by other tasks.   It looks quite nice I think:

![image](https://user-images.githubusercontent.com/208716/236613015-5523350e-5017-41d1-bd88-670622c850b7.png)

I'll work on the Jenkins file changes now, but might be pulled away again until later in the afternoon...

BeckerFrank:
@wimjongman, @ruspl-afed: Should we create a new documentation for the release process?
We can use the [Github Wiki](https://github.com/eclipse-mylyn/org.eclipse.mylyn/wiki).
Thougths?

merks:
@BeckerFrank @ruspl-afed

Please kick off a milestone build today or tomorrow morning when you're done with all the changes in preparation for m3 by the end of the day tomorrow and let me know when that's done.

My review of the current duplicate bundles in SimRel when aggregating Mylyn's current p2 repository is that Mylyn is always depending on the newer available version of that IU so that's a good news story.

merks:
[...]

PR #122, opened 2023-05-06
[#121] Prepare releng for 3.26.1

ruspl-afed:
Bulk increment to 3.26.1


wimjongman (approved):
LGTM

mylyn.github/org.eclipse.mylyn.github.core/META-INF/MANIFEST.MF
```diff
@@ -11,8 +11,8 @@ Export-Package: org.eclipse.mylyn.internal.github.core;version="6.2.0";x-friends
  org.eclipse.mylyn.internal.github.core.gist;version="6.2.0";x-friends:="org.eclipse.mylyn.github.ui",
  org.eclipse.mylyn.internal.github.core.issue;version="6.2.0";x-friends:="org.eclipse.mylyn.github.ui",
  org.eclipse.mylyn.internal.github.core.pr;version="6.2.0";x-friends:="org.eclipse.mylyn.github.ui"
-Require-Bundle: org.eclipse.mylyn.tasks.core;bundle-version="[3.26.0,4.0.0)",
- org.eclipse.mylyn.commons.net;bundle-version="[3.26.0,4.0.0)",
+Require-Bundle: org.eclipse.mylyn.tasks.core;bundle-version="[3.26.1,4.0.0)",
```
BeckerFrank, line 14: I think we need to change the Bundle-Version to 6.2.1.qualifier for all Mylyn GitHub Artifacts

BeckerFrank (changes requested):
org.eclipse.mylyn.cdt-feature version should be changed to 5.22.1
and
org.eclipse.mylyn.github.* to 6.2.1

ruspl-afed:
> org.eclipse.mylyn.cdt-feature version should be changed to 5.22.1 and org.eclipse.mylyn.github.* to 6.2.1

Great catch!

ruspl-afed:
@BeckerFrank the PR was updated to address your comments

BeckerFrank (approved):
LGTM

ruspl-afed:
Let's hold it a bit and try what Ed @merks  is [offering here](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/123#issuecomment-1537101195)

ruspl-afed:
> Let's hold it a bit and try what Ed @merks is [offering here](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/123#issuecomment-1537101195)

Ed @merks are we ready to try version increment facilities? If yes, what do we need to do so?

merks:
@ruspl-afed 

One thing I would ask and suggest is that we try to produce a milestone build from the main branch **before** you make the version increments in the main branch.

I.e., after this job finishes producing yet another nightly build, do a manual "build with parameters" and choose milestone in the drop-down:

https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/

This way we can test later that publishing another milestone with higher versioned IUs automatically deletes all previous milestone builds for the lower version IUs.  This is essential behavior for keeping the site of the sites limited (like you see how the nightly builds are limited to 5 right now)...

ruspl-afed:
> One thing I would ask and suggest is that we try to produce a milestone build from the main branch before you make the version increments in the main branch.

Done, https://download.eclipse.org/mylyn/updates/milestone/S202305081049/index.html

wimjongman:
Nice!!!

merks:
FYI, now the branding icon is there too.  Note that each simple repository has a link to the commit state of that build:

image

ruspl-afed:
Looks perfect, Ed! What is the next step here?

merks:
I don't know if you folks want to promote the milestone to a release?  I see you already have a release here so maybe that would be confusing?

https://download.eclipse.org/justj/?file=mylyn/drops/3.26.0/v20230503-2157

Of course the content of the repository in this new structure is slightly different, given the name changes as well as the additional feature that was missing.   

One advantage of publishing a release is that then this repository will exist so that downstream consumers could start using that

https://download.eclipse.org/mylyn/updates/release/latest

Note that publishing a release (doing a release build), while it does do a full build, it **does not promote** the results of that build, but rather mirrors the milestone/latest repository as the release repository.

Personally I think this is probably a good idea, but it's your call and and doing this just a matter of doing a build and choosing "release" as the build type...

ruspl-afed:
I would test the release promotion with a service release 3.26.1
As a very minimum for 3.26.1 we can complete this PR and improve our target regarding 3rd party libs,

merks:
That’s okay too. You’ll have to increment again afterwards.

merks:
FYI, you can already see the version updates in the nightly builds:

image

Issue #116, opened 2023-05-04
Support a stable repository structure for publishing builds


merks:
One approach is to reuse JustJ's repository publishing infrastructure.  I can help integrate that if that's what folks want.  But before investing effort, I want to be **sure** that's really what folks want.  I can provide details here, but for today my time to do that is limited...

Here's one example of how it would work:

https://ci.eclipse.org/emf/job/build/

The build is parameterized:

![image](https://user-images.githubusercontent.com/208716/236128178-2a65c7a1-7907-4bd5-950c-d38875594cf4.png)

The most relevant thing here is that you can specify nightly, milestone, or release and whether to promote the build.  E.g., for a PR build, one does not need to sign, probably shouldn't in fact because it slows down the build and doing that allows a contributor to grab and redistribute signed content with their "hooks" in it, and should not publish it.  

The versions are controlled by the version of a specified/designate feature so one only needs to properly increment feature versions to provide, in the end, new release versions.  One can also control which bundles/features are listed in the documentation; e.g., omit the 3rd party things.

merks:
@ruspl-afed

One problem I will have providing such support (assuming it's what folks want), is that my local build fails at the end like this:

> [INFO] org.eclipse.mylyn.tests 3.26.0-SNAPSHOT ............ FAILURE [  1.392 s]
> [INFO] Mylyn for Eclipse 4.x 3.26.0-SNAPSHOT .............. SKIPPED
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time:  02:07 min
> [INFO] Finished at: 2023-05-04T08:32:38+02:00
> [INFO] ------------------------------------------------------------------------
> [ERROR] Failed to execute goal org.codehaus.mojo:xml-maven-plugin:1.0.2:transform (default) on project org.eclipse.mylyn.tests: The directory D:\Users\test17\mylyn-main\git\org.eclipse.mylyn\org.eclipse.mylyn.tests\target\surefire-reports, which is a base directory of a ValidationSet or TransformationSet, does not exist. -> [Help 1]
> [ERROR] 
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR] 
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
[...]

PR #124, opened 2023-05-06
Support a stable repository structure for publishing builds

merks:
Fix using the wrong org.eclipse.storage.user.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/116


merks:
@ruspl-afed 

Sorry, a copy and paste error tried to use genie.cbi instead of just leaving at at the default in the pom.xml of genie.mylyn.

There is no point to this PR build because this is only a change to the Jenkinsfile which is not tested by this build.  Please merge and kick off another main branch build.

PR #126, opened 2023-05-06
Support a stable repository structure for publishing builds

merks:
Include a git commit link and removed used build.id.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/116


merks:
@BeckerFrank @ruspl-afed 

FYI, this needs to be merged to actually be tested.

Issue #128, opened 2023-05-07
Use consistent bundle and vendor names


merks:
E.g., the vendor is Eclipse Mylyn and bundle name should start Mylyn.

merks:
I think these are inconsistent as well:

![image](https://user-images.githubusercontent.com/208716/236667147-312b32d7-ec57-419a-b417-3ee0b89a73b7.png)

These have org.eclipse.mylyn namespace and appear as projects in the workspace so I think the vendor is wrong.

merks:
@ruspl-afed  @BeckerFrank 

I fixed the other ones via https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/129 but I want to be sure you agree that these should be fixed as well or if there is a reason for them to be like this...

merks:
I think a bunch of this is suspicious:

![image](https://user-images.githubusercontent.com/208716/236667895-4cc4928c-80b7-477f-a5a2-13e2413e487d.png)

Did some projects get copied from EGit?  The ones with org.eclipse.mylyn namespace are clearly provided by Mylyn but I also wonder why EGit is in their bundle names.  I know they use EGit's github API stuff, but these are not specify to EGit otherwise, right?

merks:
In any case, I'm out for about the next hour or so, so hopefully someone can provide some guidance one how best to make these consistent.

ruspl-afed:
This was long [story](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/56) and I'm not sure what was the final agreement between Mylyn and EGit regarding this.
Do you have any comments @wimjongman @BeckerFrank ?

BeckerFrank:
> This was long [story](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/56) and I'm not sure what was the final agreement between Mylyn and EGit regarding this.
> Do you have any comments @wimjongman @BeckerFrank ?

I think we should wait for @wimjongman to comment on this. I'm not sure I have the latest information.

wimjongman:
Yes. We adopted GitHub code from egit. All provider info can be changed to mylyn since we maintain it now.


Best regards,

Wim Jongman
________________________________

merks:
We're done here.  😀

PR #129, opened 2023-05-07
Use consistent bundle and vendor names

merks:
https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/128


BeckerFrank (approved):
LGTM

PR #136, opened 2023-05-08
The branding images should refer to on-line images

merks:
I mistakenly assumed the publisher copied these images, but it does not.


ruspl-afed (approved):
LGTM

ruspl-afed:
Thank you Ed @merks

Issue #131, opened 2023-05-08
Don't include test bundles and features in the update site


merks:
I thought I'd make this a separate issue so that you folks can decide if you really want the test bundles and features in the update site.

I was trying the extent to which other update sites need to be available up install the Mylyn content, but quickly noticed that the tests have dependencies that are hard to track down and that this just gets in the way of the analysis.  I suspect that probably no one wants to install tests but maybe they want them in a target platform.   But no one extends the test bundles, then I think it doesn't make sense to include these things in the update site...

merks:
Removing the things in the tests category and trying again I still get surprises like this:

![image](https://user-images.githubusercontent.com/208716/236760090-c6140df5-404d-44b2-8197-9064a20d6678.png)

This makes me think that the org.eclipse.mylyn.commons.tck feature is a test artifact...

So removing that too and I can see that all remaining Mylyn bundles and features are installable if these addition 4 repositories are available:

![image](https://user-images.githubusercontent.com/208716/236761778-bf76a9a0-3e12-42e7-b7f1-f2389c2a6b30.png)

That's also the case if there is a release train repository available, even an older one:

![image](https://user-images.githubusercontent.com/208716/236762508-048141c3-c6ba-472c-8f02-2240221b50ca.png)

BeckerFrank:
Yes, I think we can try to remove the whole "Mylyn Tests" Category from the Updatesite.
Maybe we should also remove the "Mylyn SDK and Framework" or at least remove no longer needed parts.
Thougths?

ruspl-afed:
Yes, I think we can remove tests from publishing and also fix/remove corresponding features.

PR #141, opened 2023-05-09
Remove org.eclipse.mylyn.commons.tck from the site

merks:
This appears just to have various dependencies for tests

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/131


merks:
@BeckerFrank 

It think this one should go to.  It just pulls in a bunch of things that appear are needed only for test dependencies.

ruspl-afed (approved):
Yes, this feature is full of 3rd party for testing

PR #142, opened 2023-05-12
Improve Mylyn.setup's workingsets task

merks:
Also remove the old API baseline task in favor of the new targlets.based task.

### DIFF
--- a/org.eclipse.mylyn.releng/oomph/Mylyn.setup
+++ b/org.eclipse.mylyn.releng/oomph/Mylyn.setup
@@ -126,10 +126,6 @@
         </repositoryList>
       </targlet>
     </setupTask>
-    <setupTask
-        xsi:type="pde:APIBaselineFromTargetTask"
-        name="Mylyn latest API"
-        targetName="Mylyn Modular Target for API"/>
   </setupTask>
   <setupTask
       xsi:type="setup:ResourceCreationTask"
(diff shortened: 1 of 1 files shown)


merks:
@BeckerFrank 

I tried to simplify the working set to use the Repository Predicate which now support a pattern matched against the relative path within the repository.  This avoids problems where the location might look like a regular expression..  I didn't need to exclude anything, but in case you do need that, it's helpful to use an exclusion predicate to exclude the projects in some other working set.

I also got rid of the second set baseline task because you can only really set one...

org.eclipse.mylyn.releng/oomph/Mylyn.setup
```diff
@@ -126,10 +126,6 @@
         </repositoryList>
       </targlet>
     </setupTask>
-    <setupTask
```
ruspl-afed, line 129: +1

ruspl-afed (approved):
LGTM.
Even if I don't understand some magic here.

ruspl-afed:
Let's wait for @BeckerFrank , he is The Master of Mylyn.setup

org.eclipse.mylyn.releng/oomph/Mylyn.setup
```diff
@@ -151,6 +147,7 @@
   </setupTask>
   <setupTask
       xsi:type="setup:ResourceCreationTask"
+      id="mylyn.workingsets"
```
BeckerFrank, line 150: I think that is the wrong line should this not be in line 212 instead of 150

BeckerFrank (changes requested):
Please name the correct setup Task as mylyn.workingsets.
The rest is so that we can merge.

merks:
FYI, note how setting an ID avoids positional references to the task that could be broken by textual editing:

image

BeckerFrank:
> FYI, note how setting an ID avoids positional references to the task that could be broken by textual editing:
> 
> image

Thanks I will go and add some more ID's after we have merged this Request.
The ID mylyn.workingsets  for a Resource Creation Task looks wrong for me.

merks:
Sorry, I thought I force pushed earlier.  Too many irons in the fire.  😢

BeckerFrank (approved):
LGTM

PR #144, opened 2023-05-13
all predecessor in mylyn.setup refer to tasks with ID's

BeckerFrank:


BeckerFrank:
@merks: I add ID's to all predecessor target tasks without ID's

PR #146, opened 2023-05-14
127: fix Baseline setup in Mylyn.setup

BeckerFrank:
I did not include the APIBaselineFromTargetTask in the last commit so the Baseline was not really set.

### DIFF
--- a/org.eclipse.mylyn.releng/oomph/Mylyn.setup
+++ b/org.eclipse.mylyn.releng/oomph/Mylyn.setup
@@ -126,6 +126,10 @@
         </repositoryList>
       </targlet>
     </setupTask>
+    <setupTask
+        xsi:type="pde:APIBaselineFromTargetTask"
+        name="Mylyn Baseline"
+        targetName="Mylyn Modular Target for API"/>
   </setupTask>
   <setupTask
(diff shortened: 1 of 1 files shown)


ruspl-afed (commented):
I'm not sure. How the addition should work with https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/146/files#diff-990c1078e104ec92925ec088de16bfb0fd0a93ff3313df71d28795d5ccca33e9R82 ?

BeckerFrank:
> I'm not sure. How the addition should work with https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/146/files#diff-990c1078e104ec92925ec088de16bfb0fd0a93ff3313df71d28795d5ccca33e9R82 ?

The new way to set the API Baseline is a two step.
1) define the target "Mylyn Modular Target for API"
2)  Define a Setup Task that use the newly defined target

2) was the missing Task. I can publish a new Version where I give  the new setup an ID and check if we use this as a predecessor for other tasks.

merks:
I think this is done.

ruspl-afed (approved):
Looks good

PR #148, opened 2023-05-16
Remove org.mockito.mockito-inline

merks:
This dependency isn't needed and the wrapped artifact is broken.


ruspl-afed:
Ed @merks , it was added to support mock for Java 11 +
But let's see what tests will say

merks:
This does not appear to contain content that could help anything do anything:

image

The Platform uses Mockito as well and for Java 17 without any such need...

merks:
I did some Googling and found this:

https://stackoverflow.com/questions/14292863/how-to-mock-a-final-class-with-mockito

That suggest another solution:

image

This explains the total strangeness of this bundle.

ruspl-afed:
> This does not appear to contain content that could help anything do anything:

I think it does some plugin registration for mockito that potentially can be done in other way.

But you can see that without it we have test failures:
```
Cannot mock/spy class org.eclipse.egit.core.RepositoryUtil
Mockito cannot mock/spy because :
 - final class
```

merks:
Now to actually make it work.  🙀

merks:
This comment suggested a newer version of mockito does not need this configuration:

https://www.baeldung.com/mockito-final#configure-mocktio

Locally the test worked when I used a newer version...

merks:
Two new-but-different-errors...

merks:
@ruspl-afed 

I did comment out one tiny part of one test.  Perhaps there's another way to expression that...

In any case, I really want to get rid of this nasty wrapped library that's only needed for an old version of mockito when it doesn't really make sense to stick to that old version.

So if we can get this merged, I can remove it from my the oomph simrel repo.

mylyn.tasks/org.eclipse.mylyn.tasks.ui.tests/src/org/eclipse/mylyn/internal/tasks/ui/migrator/ConnectorMigratorTest.java
```diff
@@ -424,7 +422,7 @@ public void migrateTasks() throws Exception {
 				manager.getRepositoryConnector("mock.new"), monitor);
 
 		verify(migrator, never()).getTaskData("key1", newConnector, newRepository, monitor);
-		verify(migrator, never()).createTask(argThat(not(taskData2)), any(TaskRepository.class));
+		// verify(migrator, never()).createTask(argThat(not(taskData2)), any(TaskRepository.class));
```
ruspl-afed, line 425: Ed @merks please create an issue to enable this test back and mention it here, otherwise we will forget about it.

ruspl-afed (changes requested):

mylyn.tasks/org.eclipse.mylyn.tasks.ui.tests/src/org/eclipse/mylyn/internal/tasks/ui/migrator/ConnectorMigratorTest.java
merks: Issue https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/146 created and mentioned.

merks:
@ruspl-afed 

I think this is done. 

I'd like to test that removing lucene-queries and lucene-sandbox from the target platform has any impact. No packages from either library are used in the code, so I doubt it...

merks:
@ruspl-afed 

These wrapped bundle are making me really uncomfortable that we might be making mistakes and that it will likely be very hard to correct those mistakes without some ability to version (qualifier) the wrappers:

https://github.com/merks/simrel-maven/issues/5

ruspl-afed (approved):
LGTM, Thank you Ed @merks !

merks:
I'm investigating all this lucene stuff, but for all the nice arguments about using package imports, that just seems like a bit of a disaster with these multiple lucene versions floating around.

Issue #150, opened 2023-05-17
Startup Problems


merks:
When I installed all of Mylyn in an Eclipse SDK product, I get these errors:

> org.eclipse.swt.SWTException: Widget is disposed
> 	at org.eclipse.swt.SWT.error(SWT.java:4918)
> 	at org.eclipse.swt.SWT.error(SWT.java:4833)
> 	at org.eclipse.swt.SWT.error(SWT.java:4804)
> 	at org.eclipse.swt.widgets.Widget.error(Widget.java:450)
> 	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:369)
> 	at org.eclipse.swt.widgets.Widget.getData(Widget.java:544)
> 	at org.eclipse.e4.ui.workbench.renderers.swt.ToolBarManagerRenderer.createWidget(ToolBarManagerRenderer.java:434)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.createWidget(PartRenderingEngine.java:995)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeCreateGui(PartRenderingEngine.java:659)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.safeCreateGui(PartRenderingEngine.java:763)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$2.run(PartRenderingEngine.java:728)
> 	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.createGui(PartRenderingEngine.java:712)
> 	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.subscribeChildrenHandler(PartRenderingEngine.java:312)
> 	at jdk.internal.reflect.GeneratedMethodAccessor9.invoke(Unknown Source)
> 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
> 	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:58)
[...]

PR #151, opened 2023-05-18
[#150] Startup problems

ruspl-afed:
Move ActiveReviewManager startup to happen after `APP_STARTUP_COMPLETE`

Fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/150

### DIFF
--- a/mylyn.reviews/org.eclipse.mylyn.reviews.ui/META-INF/MANIFEST.MF
+++ b/mylyn.reviews/org.eclipse.mylyn.reviews.ui/META-INF/MANIFEST.MF
@@ -8,6 +8,11 @@ Require-Bundle: com.google.guava;bundle-version="0.0.0",
  org.eclipse.core.runtime;bundle-version="0.0.0",
  org.eclipse.core.resources;bundle-version="0.0.0",
  org.eclipse.core.expressions;bundle-version="0.0.0",
+ org.eclipse.e4.core.di.annotations;bundle-version="0.0.0",
+ org.eclipse.e4.core.di.extensions;bundle-version="0.0.0",
+ org.eclipse.e4.core.services;bundle-version="0.0.0",
+ org.eclipse.e4.ui.model.workbench;bundle-version="0.0.0",
+ org.eclipse.e4.ui.workbench;bundle-version="0.0.0",
  org.eclipse.emf.ecore;bundle-version="0.0.0",
  org.eclipse.emf.ecore.xmi;bundle-version="0.0.0",
  org.eclipse.emf.edit;bundle-version="0.0.0",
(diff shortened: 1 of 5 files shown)


ruspl-afed:
> That is a lot of code for such a small issue. Is it not easier to add a workbench listener?

Do you mean [this one](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/IWorkbenchListener.java)?

wimjongman:
> > That is a lot of code for such a small issue. Is it not easier to add a workbench listener?
> 
> Do you mean [this one](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/IWorkbenchListener.java)?

Yes, then we are sure we are not too early 🤣 

I think we can just delete the code in the constructor since the window listener is already doing what the constructor does.

wimjongman:
You see that the now commented code is the same code as seen in the window listener. I can't think why this would have to be done twice.

```
	private final IWindowListener windowListener = new IWindowListener() {

		public void windowOpened(IWorkbenchWindow window) {
		}

		public void windowDeactivated(IWorkbenchWindow window) {
			window.removePageListener(pageListener);
		}

		public void windowClosed(IWorkbenchWindow window) {
		}

		public void windowActivated(IWorkbenchWindow window) {
			window.addPageListener(pageListener);
			pageListener.pageActivated(window.getActivePage());
		}
	};

	public ActiveReviewManager() {
		PlatformUI.getWorkbench().addWindowListener(windowListener);
//		IWorkbenchWindow activeWorkbenchWindow = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
//		if (activeWorkbenchWindow != null) {
//			activeWorkbenchWindow.addPageListener(pageListener);
//			windowListener.windowActivated(activeWorkbenchWindow);
//		}
	}
```

ruspl-afed:
The problem is: we start to work with Workbench before its full startup (whatever it means). 
And, I know only one way to do this: listen for `APP_STARTUP_COMPLETE` event.

* I'm not sure when `AbstractUIPlugin#start` will be invoked, but most probably before the full Workbench initialization, just because in theory it could be a part of such initialization.
* Regarding `UIJob` - it will be launched in UI thread for sure, but I'm afraid there is no relation to Workbench state.

ruspl-afed:
> That is a lot of code for such a small issue.

Hey, I added only a few lines of "code". 
All the rest is E4 magic, it is not considered! 😎

BeckerFrank:
Maybe we can use something similar to BuildsStartup (interface org.eclipse.ui.IStartup).
Thoughts?

ruspl-afed:
`IStartup` is a E2 way of what was done here with E4
It also will require a few lines of "code" and some changes for manifests.
And what will be the benefit?

BeckerFrank:
> `IStartup` is a E2 way of what was done here with E4
> It also will require a few lines of "code" and some changes for manifests.
> And what will be the benefit?

I found that we have IStartup in 3 places and did not remember that this was the old way. Sorry :-(
Now I think we should remove the two unused cases and change  BuildsStartup (the only one with logic) to the E4 way.
Should I create a new Ticket?

wimjongman:
> `IStartup` is a E2 way of what was done here with E4 It also will require a few lines of "code" and some changes for manifests. And what will be the benefit?

Yes, IStartup is not an option.

So the cause of the error is that display = null. The display is obviously not null when the UIjobs start running. 

I have tested the solution of https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/151#issuecomment-1554294145, and this works fine. So even though Alexander's solution is architecturally the best, I propose going for that simple solution.

wimjongman:
> change BuildsStartup (the only one with logic) to the E4 way.
> Should I create a new Ticket?

When we have more use for an addon, then yes, we should implement that in favor of the IStartup solution.

ruspl-afed:
> I have tested the solution of https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/151#issuecomment-1554294145, and this works fine. So even though Alexander's solution is architecturally the best, I propose going for that simple solution.

If you did so much work already @wimjongman , please propose your PR and we can close this one.

ruspl-afed:
Well, it was modern and cool, but another solution wins.

PR #152, opened 2023-05-18
Use newer Lucene bundles that with BSNs the same as the Orbit bundles

merks:


merks:
@ruspl-afed 

FYI, this is building against https://ci.eclipse.org/oomph/job/simrel/job/simrel-maven-bnd/lastSuccessfulBuild/artifact/maven-bnd/site/target/repository so is not meant to be committed.  This way I can see if the test will work before I promote such a build to the download server...

merks:
Oh yes, and of course the license checks will fail because of these pending reviews:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues

ruspl-afed:
Thank you Ed @merks

merks:
@ruspl-afed 

It looks like the IP reviews are stalled perhaps until May 25th, but I cannot wait that long to complete the changes to the changes to the https://download.eclipse.org/oomph/simrel-maven-bnd/2023-06 so we will need to go forward in parallel to the IP review process because the regular builds will fail without these changes.

Note that I'd like to clean up some of the guava dependencies and I can't do that effectively until this is merged.  In particular, these are restrict to some ancient version of guava:

image

merks:
@wimjongman 

FYI, I also need the build Mylyn night build result from this for consumption by the BIRT build that I'm working on here:

https://github.com/eclipse/birt/issues/1286

ruspl-afed:
>  so we will need to go forward in parallel to the IP review process because the regular builds will fail without these changes.

I understand, but I'm still not comfortable with failing "License vetting status check", since it may shadow some important issues.
What if we try [to configure the exclude](https://github.com/eclipse/dash-licenses/issues/110#issuecomment-938739106)  for Apache Lucene?

> Note that I'd like to clean up some of the guava dependencies and I can't do that effectively until this is merged.

How this change is related to guava?

merks:
Given previously Mylyn was already consuming the lucene-querieslucene-sandbox library and we already have this in m2, we are no more exposed now than we were.  They just fixed the license tool to do a better job checking otherwise we'd be happy assuming all the checks were done when they were not.   And because of those proper checks, you'd already be bitten by that now with the libraries already being consumed.

Note that the build failures lists the libraries:

> [INFO] p2/orbit/p2.eclipse.plugin/org.apache.lucene.queries/9.5.0.v20230521-0510
> [INFO] p2/orbit/p2.eclipse.plugin/org.apache.lucene.sandbox/9.5.0.v20230521-0510

So one "just" needs to be careful that this list doesn't grow beyond these two.

Yes, none of this makes me terribly comfortable but now the runway to m3 is growing short, parallel IP is allowed in general, and with the pressure to get this into m2 along along with the license checks not actually running properly as expected,  we've already  quite far down this path. 

Maybe other people know how to work on multiple different things in one workspace, but as long as I have a change pending committed on a branch, I don't know how to work on the next thing.

ruspl-afed (approved):
Ed @merks I really appreciate your effort and don't want to block your m3 work.

merks:
You're most welcome.  Sorry for the pressure, especially during the weekend!

Issue #138, opened 2023-05-09
Employ tycho pomless extension


ruspl-afed:
Currently we have a number of unnecessary duplications of data in our manifests that could be avoided with tycho pomless extension

Let's employ tycho pomless extension.

ruspl-afed:
This one is done, remaining manifests should be reworked individually

PR #153, opened 2023-05-19
[#138] Employ tycho pomless extension

ruspl-afed:
Remove aggregator pom.xml manifests


BeckerFrank (approved):
LGTM

ruspl-afed:
Done with aggregators. Next we need to reconfigure maven build plugins and remove pom.xml from [test]bundles where possible.

PR #155, opened 2023-05-19
[#150] Startup problems

wimjongman:
In some cases the init of the reviews plugin came to soon causing a widget disposed exception which was caused by the Display being null.

The initialization has been moved to the already existing UIJob.

### DIFF
--- a/mylyn.reviews/org.eclipse.mylyn.reviews.ui/src/org/eclipse/mylyn/internal/reviews/ui/ReviewsUiPlugin.java
+++ b/mylyn.reviews/org.eclipse.mylyn.reviews.ui/src/org/eclipse/mylyn/internal/reviews/ui/ReviewsUiPlugin.java
@@ -46,8 +46,8 @@ public class ReviewsUiPlugin extends AbstractUIPlugin {
 		//We need to schedule initialization otherwise TasksUiPlugin hasn't finished initialization.
 		UIJob job = new UIJob(Messages.ReviewsUiPlugin_Updating_task_review_mapping) {
 			@Override
-			reviewManager = new ActiveReviewManager();
 			public IStatus runInUIThread(IProgressMonitor monitor) {
+				reviewManager = new ActiveReviewManager();
 				taskReviewsMappingStore = new TaskReviewsMappingsStore(TasksUiPlugin.getTaskList(),
 						TasksUiPlugin.getRepositoryManager());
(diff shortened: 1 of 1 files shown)


ruspl-afed (approved):
This is, indeed the most laconic change that pretends to fix the startup order problem.
I would rather make `reviewManager` a final field to perform its activation in the job, because I don't like nullable fields.
But, since this is internal 'service' we can accept this change until deeper rework.

By `deeper rework` I mean turning `ActiveReviewManager` to an OSGi component, that will resolve all the startup order issues for us. Also, implementation of cascade listeners of `ActiveReviewManager` can be simplified a lot with E4.

ruspl-afed:
@wimjongman I restarted the failed pipelines to ensure that the failure was not related with your change

wimjongman:
> It could be a problem of project compilation settings
> 
> ```
> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.4:compile (default-compile) on project org.eclipse.mylyn.reviews.ui: Compilation failure: Compilation failure: 
> [ERROR] /home/jenkins/agent/workspace/pse.mylyn_150_Startup_problems_2/mylyn.reviews/org.eclipse.mylyn.reviews.ui/src/org/eclipse/mylyn/internal/reviews/ui/ReviewsUiPlugin.java:[49] 
> [ERROR] 	reviewManager = new ActiveReviewManager();
> [ERROR] 	^^^^^^^^^^^^^
> [ERROR] Syntax error on token "reviewManager", VariableDeclaratorId expected after this token
> [ERROR] 1 problem (1 error)
> [ERROR] -> [Help 1]
> ```

That plugin is still java 8. It probably does not allow accessing a private field from an anonymous class?

ruspl-afed:
We were blind. Java 8 has no guilty, the syntax was incorrect. Let's give it another chance.

wimjongman:
> We were blind. Java 8 has no guilty, the syntax was incorrect. Let's give it another chance.

Wow, that is totally weird. I confess that I used codespaces 😱 because it was a one-liner but when I pulled the changes into Eclipse I got the correct code.

image

ruspl-afed:
> but when I pulled the changes into Eclipse I got the correct code.

Ah, The Magic of Technology... I had the question: how did you manage to test this code on your side, but now I see,,,

ruspl-afed (approved):
Looks better than it was before.

Issue #159, opened 2023-05-22
Old Guava


merks:
I'm not sure what to do about com.google.guava which comes from here and is a really old version restricted to bundle-version="[21.0.0,22.0.0)"  what should we do?    We have old gson and other old things.  How best to update these things?

ruspl-afed:
I would also like to remove guava, but I'm not sure if we can do it right now.

merks:
Yes it’s work to eliminate and at best a no-op. Likely only a small highly stable subset is used such that removing version constraints is okay and expedient.

merks:
Sorry, one test I should have done before is to test installing all the Mylyn features.  This turned up a few problems, like the the rcp thing needing the old logging thing, and not all of the latest guava's dependencies being available.  I will provide a fix for that immediately.

PR #160, opened 2023-05-23
Update guava, gson, and mockito

merks:


merks:
Note that upgrading to the latest guava only ran into to errors in tests, so I do think the APIs being used are stable.  Also the gson old version is causing duplicates in SimRel so that too needs and update.  The older version of mockito was kind of pointless because other tests are using the newer version already.  I'll need to force push because I overlooked the feature.xml use of mockito...

merks:
It will be nice in the long terms to stop including 3rd party libraries in features because that locks in a specific version whereas all the actual bundle requirements are happy with a range of versions...

akurtakov:
Thanks @merks ! Guava usage has been causing all kind of issues in many places so even standardizing on one version is quite helpful.

ruspl-afed:
> It will be nice in the long terms to stop including 3rd party libraries

sure, and we are moving in this direction

merks:
@ruspl-afed 

This is done.  I'd like to do the branding changes next...

ruspl-afed (approved):
Thank you Ed @merks

PR #161, opened 2023-05-23
Ensure that 3rd party dependencies are available in the update site

merks:
This is to ensure that all of Mylyn's features install with a minimal set of other update sites available.


merks:
This change eliminates all mention of guava in the features, but includes the latest one in the site.  Also include the unfortunately required older logging version.

merks:
@ruspl-afed 

This is ready now.  Sorry I didn't get it quite right the first time.

ruspl-afed (approved):
Thank you Ed @merks

Issue #145, opened 2023-05-14
Do we want a new branding image?


merks:
Currently this is used as the feature/branding image:

![feature](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/915a35b8-9006-4e8e-a1af-e08c17a56a6a)

It looks a little dated...

For sure I'm no artist, but here's an SVG I whipped up:

![mylyn](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/7aadff7f-322a-479d-8f21-1ec8075900d6)

Here's the corresponding *.png:

![mylyn](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/80c0ffd8-4cef-4842-aacf-fd325f4c7b26)

I'm not sure it's actually better but it's more modern...

What do you guys think?

merks:
Maybe it looks better without color:

![mylyn2](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/345a7075-9f8f-44b1-8a73-f4930a7f1e66)

![mylyn2](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/d8457d71-752f-4a7c-9417-a671af660be2)

BeckerFrank:
@merks: Thank you so much for this!
You are a much better artist than me.
To me the version with the color looks nicer.
I find that the three white lines in the version without color are too dominant.

gnl42:
I think I prefer the white version, feels more like a tribute to mylyns original name, mylar.
But I think I would prefer the white more transparent.

merks:
Yes, I tend to agree that without color it looks a bit more like a transparent sheet.  With this version I used more of a gray color so that the white stripes also have some shading:

![mylyn3](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/0deabc76-ed7f-44ea-9305-e3a9eabfd04c)

![mylyn3](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/39d13bfb-e250-4357-8d56-5129969a59c1)

What do you guys think of this one?

mjmeijer:
I think the 'mylar' should be between the viewer and the eclipse logo. Now the three horizontal bars are in front.
Just having the shiny smooth triangle in front of the eclipse logo is future proof for main eclipse logo evolution

merks:
This is how it looks when it's below the stripes:
[...]

PR #162, opened 2023-05-23
Update the branding plugin images to use the new Mylyn logo

merks:
https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/145


ruspl-afed (approved):
Thank you Ed @merks

PR #163, opened 2023-05-23
Remove more includes of 3rd party libraries from features.

merks:


merks:
Note that I've tested that I can install all the features from the resulting p2 repository.

PR #164, opened 2023-05-23
include Mylyn Website in the mylyn.setup

BeckerFrank:
This is the fix for [include Website into mylyn.setup](https://github.com/eclipse-mylyn/mylyn-website/issues/4)
The Website Repository is separate installable


ruspl-afed (approved):
LGTM

PR #171, opened 2023-05-26
Fix Task Connectors working set setup

merks:


merks:
@BeckerFrank 

This fixes a small omission in the pattern.

merks:
@ruspl-afed 

Looks like dash is just failing for no good reason.  😢

merks:
It was empty so removing it is also fine, as long as there are no projects that end up in Other.  :-)

merks:
Yes, I removed it and force pushed the change.

BeckerFrank (approved):
LGTM

Issue #167, opened 2023-05-25
Many  "Not properly disposed SWT resource" in error log


jonahgraham:
When navigating around Mylyn I see the following error repeatedly with the key item on the stack seeming to be `org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.computeIconImage(DiscoveryViewer.java:549)`

<details>
<summary>full log</summary>

```

!ENTRY org.eclipse.ui.ide 4 4 2023-05-25 18:55:38.615
!MESSAGE Not properly disposed SWT resource
!STACK 0
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:655)
	at org.eclipse.jface.resource.URLImageDescriptor.createImage(URLImageDescriptor.java:311)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:290)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:268)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.computeIconImage(DiscoveryViewer.java:549)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.access$3(DiscoveryViewer.java:525)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer$ConnectorDescriptorItemUi.<init>(DiscoveryViewer.java:221)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.createDiscoveryContents(DiscoveryViewer.java:990)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.createBodyContents(DiscoveryViewer.java:611)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.discoveryUpdated(DiscoveryViewer.java:1057)
	at org.eclipse.mylyn.internal.discovery.ui.wizards.DiscoveryViewer.updateDiscovery(DiscoveryViewer.java:1538)
[...]

PR #175, opened 2023-05-28
[#167] Many "Not properly disposed SWT resource" in error log

ruspl-afed:
Move resource image resource tracking from `ConnectorDescriptorItemUi` to enclosing `DiscoveryViewer`

Fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/167

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.discovery.ui/src/org/eclipse/mylyn/internal/discovery/ui/wizards/DiscoveryViewer.java
+++ b/mylyn.commons/org.eclipse.mylyn.discovery.ui/src/org/eclipse/mylyn/internal/discovery/ui/wizards/DiscoveryViewer.java
@@ -1,13 +1,14 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2015 Tasktop Technologies and others.
- * 
+ * Copyright (c) 2009, 2023 Tasktop Technologies and others.
+ *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


wimjongman (commented):
I think you found the bug in the clearDisposables method where the disposables were cleared without being disposed of.  That was fixed. I don't get why this has to be such a big change.

There are a lot of changes in this class which are very hard to review without actually stepping through the code. Especially the modern way of using chained methods of a class instead of actual Java statements is hard to grasp.

It looks like you have replaced the "disposables" cache with an "images" cache. Is that correct?

ruspl-afed:
> I don't get why this has to be such a big change

it does not look big to me, but I may try to minimize it to the base minimum. I thought to divide it into several PRs, but in this case there is a risk that the review process will take several times longer.

> Especially the modern way of using chained methods of a class instead of actual Java statements is hard to grasp.

Well, it was never required to avoid modern Java API and prefer deeply nested `if ( x != null && x.y() != null )` blocks instead.

> It looks like you have replaced the "disposables" cache with an "images" cache. Is that correct?

Not exactly. `ImageRegistry` is used for images only, because they have natural mapping to URL and can be reused, since they have, while Fonts and other SWT resources.

wimjongman (approved):
Ok, thanks for the clarification. Let's take this as you have designed it without further rework.

ruspl-afed:
Thank you for review @wimjongman !

Issue #165, opened 2023-05-25
Features in Mylyn that aren't in SimRel


jonahgraham:
Hi folks,

I [just added](https://github.com/eclipse-packaging/packages/pull/24) most of the features of SimRel to the EPP, but I found that a bunch of features in the Mylyn update site are not being contributed to SimRel.

On a quick pass AFAICT these items don't appear in SimRel but do appear in [Mylyn](https://download.eclipse.org/mylyn/updates/milestone/S202305240755/) (as contributed to 2023-06 M3):

- org.eclipse.egit.mylyn (xref  #134)
- org.eclipse.mylyn.bugzilla_feature
- org.eclipse.mylyn.commons.activity
- org.eclipse.mylyn.github.feature
- org.eclipse.mylyn.tasks.index

I assume especially the egit and github features are or should one day be part of SimRel and the EPPs.

merks:
I added the 5 selected at the bottom (available features but not contributed to SimRel)  just now, but the other two selected were already present:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/78196c65-9ec0-4fd7-a448-65a8c88c7250)

https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/202144

jonahgraham:
Thanks Ed for the quick turnaround - I am reviewing what exactly should be in EPP based on this in https://github.com/eclipse-packaging/packages/issues/19. 

Weird thing is that `org.eclipse.mylyn.github.feature` does not appear in simrel.aggran, I would have expected it around here:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/679236/e72da69e-78ed-4324-ba9d-2e3158e67bc4)

jonahgraham:
I may be going crazy - but `org.eclipse.mylyn.github.feature` doesn't appear in SimRel output https://download.eclipse.org/staging/2023-06/

Is there a SimRel / aggregator issue here preventing it from appearing in the output?

merks:
Yes, there's something crazy going on.  Searching for the name indicates that EGit UI bundle appears to be providing that capability!

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/278db193-4a28-4365-b6f8-00104568e1c3)

@msohn

If I lock in the versions, that likely will help.  I'll get back to you in a few minutes...

jonahgraham:
[...]

PR #177, opened 2023-05-28
[#165] Features in Mylyn that aren't in SimRel

ruspl-afed:
Repackage `org.eclipse.egit.mylyn` feature to
`org.eclipse.mylyn.egit.feature`


ruspl-afed:
@jonahgraham @merks WDYT about `org.eclipse.egit.mylyn` IU rename?

jonahgraham:
IMHO you should rename bundles.

ruspl-afed:
> IMHO you should rename bundles.

do you mean "you should rename bundles as well"?

jonahgraham:
Yes.

jonahgraham:
I think this release should be treated as a new project release and now is the best time to have breaking changes. I.e. Rename and delete everything you can/want. 

FWIW I would call this release 4.0 and plan a 5.0 in the near future.

ruspl-afed:
This is too short timeframe for major release, about two weeks left for everything. The idea of upcoming 3.26.1 is to fix feedback about 3.26.0 in the timeframe of SimRel 2023-06.

Next cycle - yes, we can go with Java 17, removal of everything old, renaming, whatever else.

jonahgraham:
Btw I don't expect my opinion to carry much weight. I am mostly an interested external observer. You folk who have been resurrecting the project have my full support for whatever you deem best.

merks:
I’ll investigate tomorrow if we can rename and use p2.inf to help migrate from the older ID to the new one.

merks:
I agree that renaming the bundles and features will eliminate any confusion in the future.   I can add any p2.inf info for the feature, if possible, after this is committed.

ruspl-afed:
>  after this is committed

I have no idea when it could happen, we have a lot of committers in the project, but it is not so easy to get a review 😎

wimjongman:
I think the bundles should also be renamed. This is the first time we release github inside Mylyn, so no previous history. Let's try to start clean.

ruspl-afed:
> I think the bundles should also be renamed.

@wimjongman  they are, don't they? Or do you see other bundles that should be renamed as well? Please provide more details.

wimjongman (approved):

BeckerFrank:
> I have no idea when it could happen, we have a lot of committers in the project, but it is not so easy to get a review 😎

Unfortunately, I haven't had much time for Mylyn in the last few days, but I hope to have more time this week.

ruspl-afed:
> Btw I don't expect my opinion to carry much weight. I am mostly an interested external observer. You folk who have been resurrecting the project have my full support for whatever you deem best.

Your opinion [is welcome](https://github.com/eclipse-mylyn/org.eclipse.mylyn/discussions/193) @jonahgraham

PR #179, opened 2023-05-29
Provide a p2.inf to handle the renaming of egit.mylyn to mylyn.egit

merks:
I.e., so that org.eclipse.egit.mylyn.feature.group to can update to org.eclipse.mylyn.egit.feature.feature.group

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/165

### DIFF
--- a/mylyn.egit/org.eclipse.mylyn.egit.feature/p2.inf
+++ b/mylyn.egit/org.eclipse.mylyn.egit.feature/p2.inf
@@ -1,0 +1,6 @@
+update.id=org.eclipse.egit.mylyn.feature.group
+update.range=[0,4)
+update.matchExp = providedCapabilities.exists(pc | \
+   pc.namespace == 'org.eclipse.equinox.p2.iu' && \
+     (pc.name == 'org.eclipse.mylyn.egit.feature.feature.group' || \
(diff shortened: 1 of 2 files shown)


merks:
I've tested locally that this allows to update to the renamed feature when the installation contains the older named feature

image

image

mylyn.egit/org.eclipse.mylyn.egit.feature/p2.inf
```diff
@@ -0,0 +1,6 @@
+update.id=org.eclipse.egit.mylyn.feature.group
+update.range=[2,4)
```
ruspl-afed, line 2: I hope I understand why `4)`, but why `[2`?

merks:
The "real" IU has this:
```
<unit id='org.eclipse.egit.mylyn.feature.group' version='3.26.1.v20230523-1926' singleton='false'>
  <update id='org.eclipse.egit.mylyn.feature.group' range='[0.0.0,3.26.1.v20230523-1926)' severity='0'/>
```
So there is generally a range but indeed it could be 0 rather than 2 to deal with really really old things...

ruspl-afed:
> So there is generally a range but indeed it could be 0 rather than 2 to deal with really really old things...

I would expect either `[3` or `[0`, but may be you have some archeological inside ...

ruspl-afed (approved):
Thank you Ed @merks !

Issue #168, opened 2023-05-25
Could not load repository template extension contributed by 'org.eclipse.mylyn.bugzilla.ide' with connectorKind 'bugzilla'


jonahgraham:
In a fresh install/workspace I tried running (what will be) EPP 2023-06 M3 for C/C++ Developers and I get this error message in the log at startup:

```
!ENTRY org.eclipse.mylyn.tasks.core 4 0 2023-05-25 18:59:18.663
!MESSAGE Unexpected error while loading repository template extensions
!SUBENTRY 1 org.eclipse.mylyn.tasks.core 4 0 2023-05-25 18:59:18.663
!MESSAGE Could not load repository template extension contributed by 'org.eclipse.mylyn.bugzilla.ide' with connectorKind 'bugzilla'
```

Maybe this is a consequence of `org.eclipse.mylyn.bugzilla.ide` being included in the `ide_feature` https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/adbfdf433caeccd3e808cad3aabb37c52d447411/mylyn.context/org.eclipse.mylyn.ide-feature/feature.xml#L42 but the rest of the bugzilla feature is not installed?

merks:
This happens in the selected code when there is no registered "repKind":

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/47d95214-b256-455c-9cd3-554e551e16df)

We can see that while the bugzilla.id plugin registers the template, the bugzilla.core  registers the repository kind, so when that's not installed, we see an exception.

I think the bugzilla.ide plugin should require the bugzilla.core, and probably bugzilla.ui...

jonahgraham:
> Maybe this is a consequence of org.eclipse.mylyn.bugzilla.ide being included in the ide_feature

AFAICT from #178 the change now pulls in bugzilla all the time when using ide_feature. I think it should not pull in any of bugzilla and I have provided a fix for that in #180

BeckerFrank:
Maybe we should rework all the Entries "Mylyn SDK and Framework" because some / all are no longer needed.
Most of them were created at a time where Mylyn has many other people working on external connector and we want to provide a way tho use only parts of Mylyn.

I think now that we use Oomph for Workspace Setup we now longer really need this.

Thougths?

ruspl-afed:
I would suggest to introduce standard "YYY Developer Resources" features that contains source bundles and then remove "SDK" features.

[...]

PR #180, opened 2023-05-29
Remove bugzilla from Mylyn IDE feature

jonahgraham:
It used to be that the Mylyn IDE feature didn't really pull in any connectors, but just IDE features. However there was a log message about bugzilla as repored in #168

The current fix for #168 pulls in more of bugzilla when the ide_feature is installed.

Therefore this change is to not pull in bugzilla at all when installing IDE feature.

Fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/168


ruspl-afed:
Yes, this is good to avoid another circular dependency in Mylyn

PR #181, opened 2023-05-29
Don't transform surefire report when skipTests is true

merks:
When running with -DskipTests=true, there are no generated reports and then trying to transform them fails.


ruspl-afed (approved):
Thank you Ed @merks

Issue #170, opened 2023-05-25
Actions section of Context tab in task editor too narrow


jonahgraham:
The Actions section (see highlighted screenshot below) is too narrow to see everything and I can't see how to change the width.

This screenshot is taken on Linux using EPP for C/C++ 2023-06 M3.

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/679236/7a4efb5b-eccc-498b-9f49-d4088d42875a)

jonahgraham:
I tested on Windows too with 2023-06 M3 EPP for Java developers and it looks the same.

gnl42:
It's been like that for a long time

gnl42:
It seems to a Mylyn mindset. There are far too many hardcoded widths in the code base so you see a lot of truncated text

jonahgraham:
Thanks @BeckerFrank  - the updated layout looks better now:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/679236/703c3155-d084-450c-a0bb-0f58790672c1)

PR #183, opened 2023-05-29
Actions section of Context tab in task editor too narrow #170

BeckerFrank:
fix for UI problem (issue #170)

### DIFF
--- a/mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
+++ b/mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
@@ -231,10 +231,11 @@ public class ContextEditorFormPage extends FormPage {
 		}
 	}
 
-	private int setNewMaxWith(int oldValue, Control imageControl, Control linkControl) {
-		Point imageSize = imageControl.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
-		Point hyperlinkSize = linkControl.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
-		return imageSize.x + hyperlinkSize.x > oldValue ? imageSize.x + hyperlinkSize.x : oldValue;
+	private int calculateMaxWidth(int existing, Control image, Control link) {
+		Point imageSize = image.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
+		Point hyperlinkSize = link.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
+		int required = imageSize.x + hyperlinkSize.x;
+		return required > existing ? required : existing;
 	}
 
 	private void createActionsSection(Composite composite) {
@@ -242,7 +243,6 @@ public class ContextEditorFormPage extends FormPage {
 				ExpandableComposite.TITLE_BAR | ExpandableComposite.EXPANDED);
 		section.setText(Messages.ContextEditorFormPage_Actions);
 
-		int with = 0;
 		section.setLayout(new GridLayout());
 		GridData sectionGridData = new GridData(GridData.VERTICAL_ALIGN_BEGINNING);
 		section.setLayoutData(sectionGridData);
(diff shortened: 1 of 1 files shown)


mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
```diff
@@ -231,20 +231,27 @@ public void dispose() {
 		}
 	}
 
+	private int setNewMaxWith(int oldValue, Control imageControl, Control linkControl) {
```
ruspl-afed, line 234: It doesn't `set` anything, we can name it like `calculateMaxWidth`

ruspl-afed, line 234: Would you like to name arguments to be single word, like `existing`, 'image', `link`

mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
```diff
@@ -231,20 +231,27 @@ public void dispose() {
...
+		Point imageSize = imageControl.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
+		Point hyperlinkSize = linkControl.computeSize(SWT.DEFAULT, SWT.DEFAULT, true);
+		return imageSize.x + hyperlinkSize.x > oldValue ? imageSize.x + hyperlinkSize.x : oldValue;
```
ruspl-afed, line 237: Would you consider to extract `imageSize.x + hyperlinkSize.x` to a local variable (and inline `imageSize` and `hyperlinkSize`)?
In this case return statement will be much more readable:
```java
return required > existing ? required : existing
```

mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
```diff
@@ -231,20 +231,27 @@ public void dispose() {
...
+	}
+
 	private void createActionsSection(Composite composite) {
 		Section section = toolkit.createSection(composite,
 				ExpandableComposite.TITLE_BAR | ExpandableComposite.EXPANDED);
 		section.setText(Messages.ContextEditorFormPage_Actions);
 
+		int with = 0;
```
ruspl-afed, line 245: this is an example of `temporal coupling`
in fact what we need is to ensure that after adding all the children to `sectionClient` its `widthHint` is big enough, right? So, all this code wants to be in the very end of the block, where you do
```java
sectionClientGridData.widthHint = with
```

ruspl-afed (approved):
LGTM. Only stylistic comments.

mylyn.context/org.eclipse.mylyn.context.tasks.ui/src/org/eclipse/mylyn/internal/context/tasks/ui/editors/ContextEditorFormPage.java
merks: There is more than one link and depending on translation, it's not clear which one is wider...

Adding this to the end of the method also seemed to do the trick:
```
		sectionGridData.widthHint = section.computeSize(SWT.DEFAULT, SWT.DEFAULT).x;
```
I.e., like this on Windows (10):

image

BeckerFrank: OK, so I add some more local variables

merks: It really seems sufficient and much simpler to do only this:

image

I.e., let the section as a whole compute its preferred size...

BeckerFrank: Job with this solution is https://ci.eclipse.org/mylyn/view/GitHub%20Orga/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/view/change-requests/job/PR-183/3/

When finished we should use "Squash Merge" to bring this into main.
@merks: Thanks this indeed much less code!

Issue #184, opened 2023-05-30
Define "plugin-source" execution in the parent pom 


ruspl-afed:
Define "plugin-source" execution in the parent pom for `build` section and remove it from modules.

This is required for #138 and #182

PR #185, opened 2023-05-30
[#184] Define "plugin-source" execution in the parent pom

ruspl-afed:
* Rework tycho-source-plugin declarations for parent pom
* Remove unnecessary `tycho-source-plugin` declarations for bundles and tests
* Remove pom.xml manifests that have become "trivial" after the changes above

Fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/184


wimjongman (approved):
My kind of change: deleted lines 🎉🎉

Issue #194, opened 2023-06-13
Prepare releng to 4.0.0


ruspl-afed:
We need to prepare releng to the next release cycle:

increment version in manifests
update target platform
and so on

PR #195, opened 2023-06-13
[#194] Prepare releng to 4.0.0

ruspl-afed:
Increment versions to 4.0.0, with the following exceptions:
 * use 6.3.0 GutHub (with the hope of catching up in the near future)
 * use 5.23.0 for CDT (with the hope of catching up in the near future)


ruspl-afed:
Ed @merks , do you have any concerns regarding this PR?

ruspl-afed:
Dear Project Leads and Committers,

According to the recommended security settings _at least 1 approving review is required by reviewers with write access_ to merge a PR. Could you please find some time for review and let Eclipse Mylyn project go forward?

Thanks in advance,
AF

BeckerFrank (approved):
LGTM,
sorry I thought that you wait for an answer from Ed :-(

akurtakov:
I did a cursory look and this looks good.

merks:
Sorry. Some days I get well over a hundred emails.

I’m not so sure that consumers are well served by blanket major version increases even when nothing has changed. What can they do?  Increase their upper bound?  Increase the lower bound but get broken later by whatever change happens later to actually break API?
I will need to go through Oomph to change Mylyn bounds, some of which already span 3 major versions.  Not a huge problem but certainly questionable.

ruspl-afed:
>  that consumers are well served by blanket major version increases even when nothing has changed. What can they do? Increase their upper bound? Increase the lower bound but get broken later by whatever change happens later to actually break API?

I'm asking similar questions from time to time, and the answer I have is: [SemVer](https://semver.org). Personally I prefer to require only lower bound.

merks:
If APIs are broken I would not complain. But global search and major increment everything isn’t semantic versioning, it’s syntactic versioning or brand versioning.  Semantic versioning is doing a major increment when the semantics of a change dictates it. 

Anyway. It’s just my opinion. It don’t block the changes if folks think this is best for the project.

BeckerFrank:
> If APIs are broken I would not complain. But global search and major increment everything isn’t semantic versioning, it’s syntactic versioning or brand versioning. Semantic versioning is doing a major increment when the semantics of a change dictates it.
> 
> Anyway. It’s just my opinion. It don’t block the changes if folks think this is best for the project.

Should I open a Discussion for this?

merks:
Don’t worry. I don’t wish to belabor the point.  It’s not so easy to manage version increments one bundle at a time for a small team.

merks:
Just to close the discussion, here's what I had to change in Oomph:

https://git.eclipse.org/c/oomph/org.eclipse.oomph.git/commit/?id=1075d801d10a863181909547d774f8822da4881a

I.e., increase the upper bound to 5.0.   Then I had to minor increment the affected bundles, their exports, and all transitively including features; this is effectively  automated by errors with quick-fixes using Oomph version management tool.

ruspl-afed:
> But global search and major increment everything isn’t semantic versioning, it’s syntactic versioning or brand versioning.

I fully agree here

> It’s not so easy to manage version increments one bundle at a time for a small team.

Yes, this is just not realistic, and TBH I do not see a lot of value to invest to it with the current state of coupling between Mylyn bundles. Personally, I would support SemVer only for self-contained bundles like SWT or JFace, if ever have such a state for Mylyn we can return to SemVer best practices.

PR #198, opened 2023-06-14
fix some small problems for final eclipse 4.28

BeckerFrank:
The following was changed

- mylyn-e4.28.target
4.28-I-builds/I20230601-1220 no longer exists we now use 4.28/R-4.28-202306050440
- Mylyn.setup
use mylyn-e4.28.target as the target platform instead of mylyn-e4.26.target
- add .polyglot.* to gitignore


merks:
This looks good and is necessary.

Issue #156, opened 2023-05-19
remove IStartup Interface


BeckerFrank:
In mylyn we have three IStartup implementations. Two of the three are unused and we could delete them.
We can convert the used BuildsStartup to an other implementation.

PR #201, opened 2023-06-21
Use an OSGi declarative service instead of IStartup #156

BeckerFrank:
Replace the IStartup and org.eclipse.ui.startup extension
point by a declarative OSGi service component listening on
UIEvents.UILifeCycle.APP_STARTUP_COMPLETE.


merks:
Those look like bogus dash failures.  😢

ruspl-afed:
There are still failures in dash

BeckerFrank:
From the log:

[INFO] License information could not be automatically verified for the following content:
[INFO] 
[INFO] p2/orbit/p2.eclipse.plugin/org.eclipse.launchbar.core/2.5.100.202302061541
[INFO] p2/orbit/p2.eclipse.plugin/org.eclipse.launchbar.ui/2.5.100.202303140100
[INFO] p2/orbit/p2.eclipse.plugin/org.eclipse.tools.templates.core/1.3.0.202211062329
[INFO] p2/orbit/p2.eclipse.plugin/org.eclipse.tools.templates.freemarker/1.3.0.202211062329
[INFO] p2/orbit/p2.eclipse.plugin/org.eclipse.tools.templates.ui/1.4.0.202211062329
[INFO] 
[INFO] This content is either not correctly mapped by the system, or requires review.

but I only found references to org.eclipse.launchbar or org.eclipse.tools.templates.freemarker in old target platform files.

- mylyn-e4.6.target
- mylyn-e4.7.target
- mylyn-e4.8.target
- mylyn-emaintenance.target
- mylyn-estaging.target

ruspl-afed:
Actually I have no idea now to procced further with broken Dash.
The project activity is actually blocked.
I suggest to disable Dash action until we can get reply from IP team. 
WDYT @BeckerFrank @wimjongman @merks ?

merks:
I run builds with a parameter to disable failure conditionally:
```
-Ddash.fail=$DASH_STRICT
```
So I believe that using `-Ddash.fail=false` will still do the checking, still report on the problems, and still open new issues as needed, but will not fail the build.   I'm not sure about the opening CQ thing in this case; I run dash as a regular part of the build for my Orbit-like stuff...

I believe @waynebeaton recommends not failing builds in such a way...

The above issues actually look a bogus hit.  Issues requiring manual intervention are taking a bit longer latest to clear the backlog the last while...

ruspl-afed:
> and still open new issues as needed

are you sure regarding this? I thought one need to do it explicitly via `/request-license-review`.

> I believe @waynebeaton recommends not failing builds in such a way...

Yes, true, he does. My concern is that if it doesn't fail, no one will take care to view the log of a check that didn't fail.

merks:
It's a legitimate concern.  This is why I parameterize my build, but that doesn't help here.   So you don't have many alternatives:  Wait for goodness knows how long, or risk not noticing ip issues...

ruspl-afed:
@BeckerFrank .target removals seems to be unrelated

mylyn.context/org.eclipse.mylyn.team.ui/plugin.xml
```diff
@@ -17,14 +17,15 @@
    <extension-point id="commitTemplates" name="Commit Template Handlers" schema="schema/commitTemplates.exsd"/>
     
     <!-- needed to update change sets -->
+<!--
```
ruspl-afed, line 20: Let's just remove this extension if we don't need it

ruspl-afed (changes requested):
Please consider removing commented extensions

mylyn.context/org.eclipse.mylyn.team.ui/plugin.xml
BeckerFrank: fixed with new commit

mylyn.tasks/org.eclipse.mylyn.tasks.ui/plugin.xml
```diff
@@ -56,14 +56,14 @@
         <category name="%newWizards.category.name" id="org.eclipse.mylyn.ui.wizards.new"/>
     </extension>
 
-    <extension
+<!--    <extension
```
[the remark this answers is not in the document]
BeckerFrank: fixed with new commit

ruspl-afed (approved):

Issue #182, opened 2023-05-29
Introduce standard "YYY Developer Resources" features that contains source bundles


ruspl-afed:
As discussed [here](https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/168#issuecomment-1567255060) we need to introduce "YYY Developer Resources" features and then remove "SDK" features.

* [x] `org.eclipse.mylyn.builds.sdk` -> `org.eclipse.mylyn.builds.feature.source`
* [x] `org.eclipse.mylyn.commons.sdk` -> `org.eclipse.mylyn.commons.feature.source`
* [x] `org.eclipse.mylyn.context.sdk` -> `org.eclipse.mylyn.context.feature.source`
* [x] `org.eclipse.mylyn.reviews.sdk.feature` -> `org.eclipse.mylyn.reviews.feature.source`
* [x] `org.eclipse.mylyn.sdk_feature` -> `org.eclipse.mylyn.tasks.feature.source`
* [x] `org.eclipse.mylyn.versions.sdk` -> `org.eclipse.mylyn.versions.feature.source`

ruspl-afed:
I would like to do this together with revising feature's content/ids. So, this effort doesn't fit 3.26.1 service release.

jonahgraham:
A new +1 on this - the current SDK bundle has tight requirements on specific Orbit produced third-party bundles. While the MANIFEST.MF's do Import-Package. xref: https://github.com/eclipse/orbit/pull/37

ruspl-afed:
Ed @merks it seems that bundle `org.eclipse.mylyn.commons.net` was included only to SDK feature and now we need to include it to another feature explicitly. I plan to do this tomorrow morning with your review support. If you think that something else should be done before M1 please mention it here.

merks:
Very good. I’m here to help in the morning.

ruspl-afed:
Found another orphan `org.eclipse.mylyn.commons.identity.ui`. Need to check all of them.

merks:
I just started looking as well...

ruspl-afed:
I summarized my findings using #240

ruspl-afed:
Fixed with #238 #240

Issue #174, opened 2023-05-27
Replace references to "Hudson" with "Jenkins" in the Hudson UI


gnl42:
As discussed in #63, replace references to Hudson With Jenkins that are visible to the user in the Hudson UI.

Leave any API changes for another time

gnl42:
How do I satisfy the attribution requirement of https://www.jenkins.io/artwork/ for the jenkins icon?

gnl42:
Done in #63

PR #203, opened 2023-07-02
first step for #182 (restructure Updatesite)

BeckerFrank:
Here how the new site is structured!
image
image


merks (approved):
Looks good to me.

One thing this reminds me is that the JustJ update site publisher has support for removing categories during the publishing process.   Maybe we should use that to remove this Uncategorized category to make it even harder for users to accidentally install things in there?

ruspl-afed:
What was inside "Uncategorized" @BeckerFrank ?

merks:
The repository explorer currently shows this:

image

I.e., 3rd party dependencies that you want in the update site but are not included in any features included in the update site.

BeckerFrank:
> Looks good to me.
> 
> One thing this reminds me is that the JustJ update site publisher has support for removing categories during the publishing process. Maybe we should use that to remove this Uncategorized category to make it even harder for users to accidentally install things in there?

+1 for remove the Uncategorized category.
Should I merge this Pull Request and you continue with the remove of the Uncategorized category in a new Pull Request?

merks:
Yes that’s a separate issue so let’s do that separately.

ruspl-afed:
Cool. Only 2 questions:
1) How improvement of branding helps with category.xml?
2) What about failing tests?

BeckerFrank:
> Cool. Only 2 questions:
> 
> 1. How improvement of branding helps with category.xml?

We had the same name for to IU that I fixed by add a branding for one of them.

> 2. What about failing tests?

For some reason the Hudson and Jenkins default services were not available behind https://mylyn.org.
I think I am now the only person who knows the secrets of the server and can fix this.

We should create a new instance and hope that the tasktop sponsored one works until the new instance is set up.

gnl42:
We did decide to drop support for Hudson (#174) so that simplifies things a bit.

Can we use a docker image for the tests, remove the need for a "sponsored" server?

ruspl-afed (approved):

BeckerFrank:
> We did decide to drop support for Hudson (#174) so that simplifies things a bit.
> 
> Can we use a docker image for the tests, remove the need for a "sponsored" server?

Yes we have docker images for Jenkins and Bugzilla (see https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/org.eclipse.mylyn.releng/multipass/roles/jenkins and https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/org.eclipse.mylyn.releng/multipass/roles/bugzilla).

But how we can use a docker container in the pipeline for tests?

gnl42:
So I've come up with a working PoC using TestContainer to launch Jenkins as docker container and using that container to run a couple of of JUnit tests successfully. 

It runs fine under Eclipse but when I try build Mylyn with Maven the build blows on the modified target file.
`[ERROR] Failed to resolve target definition file:/E:/workspaces/eclipse-mylyn/mylyn-main/git/org.eclipse.mylyn/org.eclipse.mylyn-target/mylyn-e4.28.target: Artifact MavenArtifactFacade [...], IncludeDependencyScope = [compile, runtime, test], MissingManifestStrategy = GENERATE, IncludeSource = true is not a valid jar file`

Also I don't know if it will run using the Mylyn Jenkins instance?

ruspl-afed:
Dear @gnl42 
Thank you for your effort to improve Jenkins integration test infrastructure.
Please extract the thread regarding Jenkins to a dedicated issue. 
This MR had another focus and already merged, no further discussions are expected here.

PR #239, opened 2023-07-11
Exclude the Uncategorized category from the promoted update site

merks:


merks:
The promoted update site will only contain the categorized items like in this local build:

image

ruspl-afed (approved):
LGTM

BeckerFrank:
> The promoted update site will only contain the categorized items like in this local build:
> 
@merks: how do you build this local in /org.eclipse.mylyn-site/updates/nightly ? I have used the one in target/repository but this includes the Uncategorized category.

merks:
I have an external tools launch configuration for kicking off a Maven build:

image


I can even point it at a local repository of a locally build JustJ publisher, but you don't need that.

The important three parameters are selected, i.e., one to enable the profile and the other to specify where to publish the results; in this case in the site project, but not in the target folder because that folder is deleted each build but the site accumulates the results and I can test doing milestone and release builds.

Issue #241, opened 2023-07-13
Ensure that Mylyn update site is complete


merks:
This is a follow up on these failures

https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/203130

The next failure is this one:

```
JavaPackage(org.apache.ws.commons.serialize [1.0.2,1.1.0)) is required by:
  ValidationSet(main)
    Contribution(Mylyn)
      MappedRepository(file:/D:/Users/merks/mylyn-main/git/org.eclipse.mylyn/org.eclipse.mylyn-site/target/repository)
        Feature(org.eclipse.mylyn.bugzilla.feature.feature.group [4.0.0.v20230713-0631,4.0.0.v20230713-0631])
          InstallableUnit(org.eclipse.mylyn.bugzilla.core 4.0.0.v20230616-0701)
            InstallableUnit(org.apache.xmlrpc.common 3.1.3.v20181004-1955)
```

merks:
@ruspl-afed

I will need to add things and retest until there is no failure.

merks:
This next problem is more problematic because it looks like DLTK is include Mylyn's feature but Mylyn has changed the ID.
```
InstallableUnit(org.eclipse.mylyn.context_feature.feature.group 3.0.0) is required by:
  ValidationSet(main)
    Contribution(PDT)
      MappedRepository(https://download.eclipse.org/tools/pdt/updates/8.0/)
        Feature(org.eclipse.php.mylyn.feature.group)
          InstallableUnit(org.eclipse.dltk.mylyn.feature.group 6.3.0.202305301153)
```
I think I need to disable that feature....

merks:
After disabling the DLTK and PDT feature, then this:
```
JavaPackage(com.google.gerrit.common.auth 2.1.5) is required by:
  ValidationSet(main)
    Contribution(Mylyn)
      MappedRepository(file:/D:/Users/merks/mylyn-main/git/org.eclipse.mylyn/org.eclipse.mylyn-site/target/repository)
        Feature(org.eclipse.mylyn.gerrit.dashboard.feature.feature.group [4.0.0.v20230616-0701,4.0.0.v20230616-0701])
          InstallableUnit(org.eclipse.mylyn.gerrit.dashboard.core 4.0.0.v20230616-0701)
            InstallableUnit(org.eclipse.mylyn.gerrit.core 4.0.0.v20230616-0701)
```

PR #242, opened 2023-07-13
[#182] Add required 3rd party bundles

ruspl-afed:
* reworked "org.apache.*" dependencies from `Require-Bundle` to `Import-Package`
* added missed bundle entries to `category.xml`

Rework of other dependencies like guava is planned for other PRs


merks:
I'm not sure the plan here, but the cycle time for the testing will take way way too long.  Please have a look at

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/241

ruspl-afed:
I'll delay these changes, Ed @merks has quicker solution with #241

ruspl-afed:
I plan to merge this one when pipeline will succeed Ed @merks

merks (approved):
Note that this issue is highly relevant:

https://github.com/eclipse-orbit/orbit-simrel/issues/1

There are lots of new versions of dependencies including new versions of lucene.

I want to get rid of this too:

https://github.com/eclipse-orbit/orbit/issues/48

But some Orbit things on which Mylyn depends still explicitly use it...

None of this speaks against your changes rather that further changes will be relevant.

PR #249, opened 2023-07-18
[#248] No explicit encoding in Mylyn projects

### LINKED ISSUE #248: No explicit encoding in Mylyn projects
Currently there are 153 projects with no explicit encoding set. It would be good to set UTF-8 encoding in each ```.settings```.

sovesti:
Explicitly defined encoding in all projects.

Closes #248


ruspl-afed (approved):
LGTM. Thanks @sovesti !

Issue #254, opened 2023-07-20
java.lang.Boolean constructor is deprecated


sovesti:
There are several uses of [deprecated](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/Boolean.html) ```Boolean(boolean value)``` constructor in Mylyn codebase. It should be replaced with ```Boolean.valueOf(boolean value)```.

gnl42:
Can we just get the Java17 code cleanup/format conventions (#200) settled on and do a global cleanup?

BeckerFrank:
@ruspl-afed: Can we close this ticket? Or is something left?

ruspl-afed:
@sovesti are you done here? If so, please close the ticket.

PR #255, opened 2023-07-20
[#254] java.lang.Boolean constructor is deprecated

sovesti:
Replaced all occurences of ```Boolean(boolean)``` with
```valueOf(boolean)```


gnl42:
Why don't you let java autobox the conversion?

sovesti:
@gnl42 thank you very much! I didn't know that Eclipse supports that kind of clean ups. I've applied "valueOf() rather than instantiation" clean up to all resources and updated the PR.

ruspl-afed (approved):
LGTM

Issue #200, opened 2023-06-16
Create Java->Code Style configuration for Java 17 and other best practices


gnl42:
The Mylyn workspace is using the default Eclipse 'Code Style' settings for 'Code Cleanup' and 'Formatting' (which seem to be stuck in the JavaSE-1.5 era ???)

With Java17 now the default for the project there are a lot of settings that we can take advantage of to create cleaner code.

gnl42:
I exported CleanUp and Formatter  locally but GitHub won't allow me to add to the issue *&^%^(

I don't know the Eclipse internals enough as how to set up the projects to use them though.

@BeckerFrank 
Not sure where to go from here??

BeckerFrank:
> I exported CleanUp and Formatter locally but GitHub won't allow me to add to the issue *&^%^(
> 
> I don't know the Eclipse internals enough as how to set up the projects to use them though.
> 
> @BeckerFrank Not sure where to go from here??

I think we have the settings on the single project in the .settings folder.
Can you place the exported file in the project org.eclipse.mylyn.releng so that I can try to import the file and look how we can set it for all projects.

merks:
Are you trying to create project-specific settings?  That’s a lot of copying to manage! 

This can also be set globally but then we would want to do that in the Oomph setup so that we all have it configured the same way. 

Will this change the save options such that any file we touch will be formatted with new formatting?

merks:
Project-specific settings are best because there is no special setup required. But keeping them consistent is not trivial.

BeckerFrank:
I think we can use the following.

-  use one project as the source and set here the wanted local settings for CleanUp and Formatter
- create a script that replace the files org.eclipse.jdt.ui.prefs and org.eclipse.jdt.core.prefs in all java projects
- if needed we can instead of the replace only do a delete and insert of the  CleanUp and Formatter Settings 

For new Java projects we must copy the two files to the new project.

Thoughts?

ruspl-afed:
+1 for project specific settings. 
Yes, they require more effort to sync but guarantees that changes done in any Eclipse IDE will use the same settings. I know that some people create linked .pref file that points to a common file, but I have no practical experience with this solution.

@gnl42 it is a bit tricky to discuss the change in this form.
Could you please highlight somehow the difference between current and suggested settings? Perhaps you can create a PR for one project to let us see what exactly we try to introduce.

BeckerFrank:
I add the defaults as a separate files.

[...]

PR #259, opened 2023-07-23
Code Style for Java 17 #200

gnl42:
Restarted the Code Style branch with just one one project, org.mylyn.commons.core:

Main highlights:

- Take advantage of as many Java 17 features as possible (that makes sense)
   - Spaces for tabs
   - boxing/unboxing
   - lambdas cleanup
   - simplify expressions when possible 
- Make autosave cleanup match above
- Updated the 'New Class' copyright text. References epl-v20 now

@BeckerFrank I'm not sure how you meant to use the CodeStyleTemplates folder that you created in the original branch.
I don't see any references to it so I guess there is still the issue of how apply the settings to all the project (and new projects) and keep them in sync with any changes?


BeckerFrank:
> Restarted the Code Style branch with just one one project, org.mylyn.commons.core:
> 
> Main highlights:
> 
> * Take advantage of as many Java 17 features as possible (that makes sense)
>   
>   * Spaces for tabs
>   * boxing/unboxing
>   * lambdas cleanup
>   * simplify expressions when possible
> * Make autosave cleanup match above
> * Updated the 'New Class' copyright text. References epl-v20 now
> 
> @BeckerFrank I'm not sure how you meant to use the CodeStyleTemplates folder that you created in the original branch. I don't see any references to it so I guess there is still the issue of how apply the settings to all the project (and new projects) and keep them in sync with any changes?

I just want to have a single point of trust for the two files and an easy way to copy the files as needed for new projects. Maybe have a regular schedule to check that there are no discrepancies.

@wimjongman @ruspl-afed @merks Thougths?

merks:
At one point I prototyped a tool in Oomph for managing project-specific preferences, but I never had time to complete it, though it was pretty far along. The idea was that some project could be designated as the source of a particular set of preferences and if those preferences are changed in that project, it would propagate automatically to the other projects (and if you changed some downstream project inconsistently, it would tell you).  It would tell you in general about any project-specific preference that was "unmanaged".  Alas there isn't time for everything.  Everyone suffers from this problem...

gnl42:
How about for now we manually modify all the project setting files and deal with the sync problem after that?

mylyn.commons/org.eclipse.mylyn.commons.core/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -1,15 +1,158 @@
...
+cleanup.instanceof=true
+cleanup.instanceof_keyword=true
+cleanup.invert_equals=false
+cleanup.join=true
+cleanup.lazy_logical_operator=false
+cleanup.make_local_variable_final=true
+cleanup.make_parameters_final=true
+cleanup.make_private_fields_final=true
+cleanup.make_type_abstract_if_missing_method=false
+cleanup.make_variable_declarations_final=true
```
ruspl-afed, line 51: Does it mean that we will have `final` keyword everywhere?

mylyn.commons/org.eclipse.mylyn.commons.core/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -345,20 +426,67 @@ org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_method_decla
...
+org.eclipse.jdt.core.formatter.parentheses_positions_in_if_while_statement=common_lines
+org.eclipse.jdt.core.formatter.parentheses_positions_in_lambda_declaration=common_lines
+org.eclipse.jdt.core.formatter.parentheses_positions_in_method_delcaration=common_lines
+org.eclipse.jdt.core.formatter.parentheses_positions_in_method_invocation=preserve_positions
+org.eclipse.jdt.core.formatter.parentheses_positions_in_record_declaration=common_lines
+org.eclipse.jdt.core.formatter.parentheses_positions_in_switch_statement=common_lines
+org.eclipse.jdt.core.formatter.parentheses_positions_in_try_clause=common_lines
 org.eclipse.jdt.core.formatter.put_empty_statement_on_new_line=true
-org.eclipse.jdt.core.formatter.tabulation.char=tab
+org.eclipse.jdt.core.formatter.tabulation.char=space
```
ruspl-afed, line 473: AFAIU, this will change the whole source base. What is our benefit?

ruspl-afed:
@gnl42 there are two main concerns from my side:
1) converting all the tabs to spaces
2) adding final keywords everywhere

gnl42:
@ruspl-afed 

1. My take on tabs is that depending on the client used to work/view the code, is that the indentation can vary, distorting the appearance of the layout. The only argument I can think of for keeping them is that they make the file slightly smaller but disk is cheap so it's not a strong argument as far as I'm concerned.
2. I prefer to let the compiler do as much as work as possible preventing accidental typos. `final` to me makes it clearer what the intent of a variable is and also serves as a visible reminder of what is going on. Plus, they can allow the compiler to produce faster code.

What are your objections to these two changes?

ruspl-afed:
@gnl42 
1. Personally, I don't prefer either a space or a tab for a new code base. However, for Eclipse-based sources I would tune format as close to Eclipse Platform as possible. For Eclipse Mylyn code base my argument is to not needlessly mess up the history of the code. I would postpone tab to space change in any case, to not mix it with other changes that will be introduced by activating clean-ups.

2. Regarding final. Compiler already considers variables final by default, so I don't think we need to force excessive keyword everywhere.
> starting in Java SE 8, a local class can access local variables and parameters of the enclosing block that are final or effectively final. A variable or parameter whose value is never changed after it is initialized is effectively final.

 https://docs.oracle.com/javase/tutorial/java/javaOO/localclasses.html#accessing-members-of-an-enclosing-class

gnl42:
@ruspl-afed 

1. Can you clarify what you mean by 'messing up history'? It doesn't affect git history and, assuming the diff client is configured to ignore white space changes, only code changes will show up when comparing versions.
2. Real use of `final ` is prevent coding errors, `effectively final` doesn't do that

merks:
I agree with @ruspl-afed  concerns.  I see it often in the platform that when you show revision annotations that you see the last time someone changed the format and not the history of meaningful semantic changes.

gnl42:
I've switched indentation back to tabs and removed the use of `final `

Not sure why github seems to be saying `final `is still enabled but my local copy has it disabled.

ruspl-afed (approved):
Let's try it out

PR #262, opened 2023-07-28
Update plugin.xml

BeckerFrank:
fix regression from [#165] Features in Mylyn that aren't in SimRel


merks (approved):
Oh dear. I wonder why that’s not flagged an error in the workspace.

Issue #250, opened 2023-07-19
Rework 3rd party libs dependencies


ruspl-afed:
* switch from `Require-Bundle` to `Import-Package`
* ensure that we don't have 3rd party bundles referenced in `feature.xml`
* ensure that we do publish required 3rd party bundles in `category.xml`

PR #263, opened 2023-07-29
[#250] Rework 3rd party libs dependencies

ruspl-afed:
Rework guava access from `Require-Bundle` to `Import-Package`


merks (approved):
Looks fine.

PR #264, opened 2023-07-29
[#250] Rework 3rd party libs dependencies

ruspl-afed:
Rework guava access from `Require-Bundle` to `Import-Package`

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.sdk.util/META-INF/MANIFEST.MF
+++ b/mylyn.commons/org.eclipse.mylyn.commons.sdk.util/META-INF/MANIFEST.MF
@@ -4,8 +4,7 @@ Bundle-Name: Mylyn Commons SDK Utilities
 Bundle-SymbolicName: org.eclipse.mylyn.commons.sdk.util;singleton:=true
 Bundle-Version: 4.0.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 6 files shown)


mylyn.commons/org.eclipse.mylyn.commons.notifications.tests/META-INF/MANIFEST.MF
```diff
@@ -14,4 +13,5 @@ Require-Bundle: com.sun.xml.bind;bundle-version="2.3.3",
 Export-Package: org.eclipse.mylyn.commons.notifications.tests;x-internal:=true,
  org.eclipse.mylyn.commons.notifications.tests.core;x-internal:=true,
  org.eclipse.mylyn.commons.notifications.tests.feed;x-internal:=true
+Import-Package: com.sun.xml.bind;version="2.3.3"
```
merks, line 16: I haven’t checked but it would be good it the was not needed at all.

ruspl-afed: I'll have a look how we can eliminate it

ruspl-afed: I'll create another PR to try luck with `com.sun.xml.bind` @merks

Issue #260, opened 2023-07-23
Changes to the 'Commit Comment Template' are not reflected in the commit message


gnl42:
Changes to the 'Commit Comment Template' isn't being honoured by the git commit message box. 

I changed the template to say:
`${task.description} #${task.key}

Task-Url:  ${task.url}`

but the commit message still used the default:
`${task.key}: ${task.description}

Task-Url: ${task.url}`

BeckerFrank:
Are you sure that you changed the right template.
We have a global template in Settings -> Mylyn -> Team and you can overwrite this in a project with properties -> Task Repository -> Commit Template

gnl42:
Sigh, I didn't know about the project setting. Kinda hidden

Can someone please explain the philosophy behind the setup?
We have a dedicated workspace for Mylyn, yet every project has it's own project settings overriding the workspace defaults so we end up with something like #265 when a global setting needs to be changed

BeckerFrank:
> Sigh, I didn't know about the project setting. Kinda hidden
> 
> Can someone please explain the philosophy behind the setup? We have a dedicated workspace for Mylyn, yet every project has it's own project settings overriding the workspace defaults so we end up with something like #265 when a global setting needs to be changed

We have the settings in all projects before we had a workspace for Mylyn. We want to make sure that all Mylyn projects use the same settings and don't want the user to have to make a change to their global settings.

Using the Eclipse installer Mylyn.setup, we can now set the global settings and remove the project specific settings. How do we make sure that each committer / contributor uses the Mylyn.setup

gnl42:
> How do we make sure that each committer / contributor uses the Mylyn.setup

Peer pressure? Don't approve PR?

PR #274, opened 2023-08-04
switch to global commit template #260

BeckerFrank:
Here the version with the global setting (coming from mylyn.setup) and the4 deleted local settings.

@merks: Is it correct that we we use a User Preferences without an Annotation that the user can only temporary change the Preference and the next start we again set it to the value stored in the mylyn.setup


merks (approved):
The annotation and compound tasks just help the preference recorder record things in a sensible common place, but they have no semantic significance.

A user could always override a preference task in their user.setup, workspace.setup, or installation.setup if they wish.  But in this case we'd all like to use the same common preference in a Mylyn workspace, so have that in the Mylyn setup in one simple place is ideal.

PR #284, opened 2023-09-19
Fix urls in target

BeckerFrank:
The following urls are no longer valid.

- https://download.eclipse.org/eclipse/updates/4.29-I-builds/I20230726-1800
- https://download.eclipse.org/tools/cdt/builds/11.3/cdt-11.3.0-m1

Also we have a issue with org.hamcrest.Factory so we need to change more code.


ruspl-afed:
Hi @BeckerFrank , it seems that similar things was already done with #283

BeckerFrank:
> Hi @BeckerFrank , it seems that similar things was already done with #283

Yes this is no longer needed. The only diff is that in my Version we have  org.hamcrest.core;bundle-version="[1.0.0,2.0.0)" and in the other one this line is removed.

ruspl-afed:
> and in the other one this line is removed.

Yes, the removal was a kind of brave experiment from my side. I hope we will be able to remove some older lib versions as a result.

Issue #287, opened 2023-09-29
Remove guava


gnl42:
Replace guava methods with native java methods

PR #302, opened 2023-11-01
[#287] Remove guava

ruspl-afed:
Move newly introduced `LinkedHashMappArrayListValuedHashMap` to a provisional API package.

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
+++ b/mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
@@ -6,6 +6,7 @@ Bundle-Version: 4.1.0.qualifier
 Bundle-Vendor: %Bundle-Vendor
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Export-Package: org.eclipse.mylyn.commons.core,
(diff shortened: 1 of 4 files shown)


ruspl-afed:
I hope this change is fine with you @gnl42

gnl42:
I'm good with it @ruspl-afed.

I don't really understand how the packages are laid out.
So often I see 'Discouraged access: XXX is not API' for classes that obviously are needed by the connectors.
If they are not meant for public consumption, why are the 'required' to make the connectors work?

ruspl-afed:
The API package/type/member is the subject of LTS, this is a kind of our public commitment to Mylyn consumers.
Although we can change API via "major release" procedure, usually people are not happy with API changes, since it mean more work for them as well.

So we should think very carefully before declaring a type to be public Mylyn API.
In contrast with this, "internal API" could be 'required' for connectors to work, but we are still free to improve it even with "breaking" changes when we need.

Usual package layout practice is too put proposed API to non-internal package, but mark it as "internal" for several releases to see how it works for consumers.

Regarding this very type I have additional concerns since it exposes 3rd party API to consumers, I would prefer to declare an interface here and then populate it with the proposed type as one of the possible implementations.

Issue #301, opened 2023-10-29
Upgrade Tycho to version 4.0.3


gnl42:
Tycho 4.0.0 requires Maven 3.9.x which is the default for Jenkins so that requirement is met.

BeckerFrank:
[jenkinsbuild 155](https://ci.eclipse.org/mylyn/view/GitHub%20Orga/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/155/) is the first run with the new version

akurtakov:
Meanwhile Tycho 4.0.4 is out https://projects.eclipse.org/projects/technology.tycho/releases/4.0.4

PR #308, opened 2023-11-03
[#301] update to Tycho 4.0.3

BeckerFrank:


gnl42:
I gave it a quick try initially but gave up because the "Problems" view showed some 5/6 complaints about options not being valid and I had no idea what the alternatives would be.

wimjongman:
> I gave it a quick try initially but gave up because the "Problems" view showed some 5/6 complaints about options not being valid and I had no idea what the alternatives would be.
> 

Can you attach the text behind the lines?

gnl42:
image

wimjongman:
I mean, if you double-click on a line, is there more information?

BeckerFrank:
I think it's some kind of caching problem. I delete all warnings and then force a rebuild with Clean... after that I can't see the warnings anymore.
The warning appears again when I switch back to the main branch. The same fix works here as before.

gnl42:
I guess you're right about it being some sort of caching problem.

When I noticed the warnings  I switched back to 3.0.3 and they went away. Now when I tried it they remained ???

I had a quick look at the official documentation for some of the plugins and the options are not listed. I guess they are remnants of the early days of the tycho plugins and can be removed

gnl42 (approved):
LGTM

Issue #313, opened 2023-11-16
Move Mylyn Doc Task Editor Integration to the Mylyn Repository


BeckerFrank:
From now on we will have the Task Editor Integration no longer in the Mylyn Docs Repository. 

Details can be found in https://github.com/eclipse-mylyn/org.eclipse.mylyn.docs/issues/131

ruspl-afed:
Now Mylyn Docs is a part of Mylyn repository

PR #314, opened 2023-11-18
[#313] move MylynDocs org.eclipse.mylyn.wikitext.tasks.ui 

BeckerFrank:
[#313] move MylynDocs org.eclipse.mylyn.wikitext.tasks.ui Project to Mylyn as org.eclipse.mylyn.tasks.wikitext.ui

(actual commit was https://github.com/eclipse-mylyn/org.eclipse.mylyn.docs/commit/61491d1dafa81eff210be220585822e06c5e8cb7)


org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
ruspl-afed, line 1: Please reduce diff for .target @BeckerFrank , it is hard to review in this state

gnl42:
Could you add org.eclipse.mylyn.commons.core/.settings/{org.eclipse.jdt.core.prefs,org.eclipse.jdt.ui.prefs} to the .settings folder

gnl42:
Did you copy/past the files or merge repositories? 
I don't see any git history for the new files

merks:
I still get the sense that it would be in the long terms interest to merge the repos as was already done with all the separate repos....

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
BeckerFrank: Sorry, this is caused by different line endings.

I think we should create a .gitattributes file to prevent this from happening in the future.

I have added line 87 to 90

BeckerFrank:
> Could you add org.eclipse.mylyn.commons.core/.settings/{org.eclipse.jdt.core.prefs,org.eclipse.jdt.ui.prefs} to the .settings folder

I have copied the .settings folder from org.eclipse.mylyn.gitlab.core. This folder only contains the org.eclipse.jdt.core.prefs. So I have to create the org.eclipse.jdt.ui.prefs also in the gitlab projects.

BeckerFrank:
> Did you copy/past the files or merge repositories? I don't see any git history for the new files

Yes because I only want to move one project.

BeckerFrank:
> I still get the sense that it would be in the long terms interest to merge the repos as was already done with all the separate repos....

Yes, we can start a discussion with @akurtakov (the Mylyn Docs committer).

If we merge repositories, we should answer the following questions.

- Is it ok that we have the same release dates?
- Can we have a single Jenkins build
- Can we set up a single configuration of .github workflows?
- Do we still need two separate p2 update sites?
- maybe some more ...

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
BeckerFrank: I now have added

```shell
core.autocrlf=input
``` 
to my global git config and this should solve my problem.

akurtakov:
> Yes, we can start a discussion with @akurtakov (the Mylyn Docs committer).
> 
> If we merge repositories, we should answer the following questions.
> 
>     * Is it ok that we have the same release dates?

Assuming Mylyn releases regularly, this sounds ok.

> 
>     * Can we have a single Jenkins build

Yes, it's possible but one would have to work quite a bit to make it happen. IMHO if this doesn't happen there is no point in merging.

> 
>     * Can we set up a single configuration of .github workflows?

I don't see an issue here.

> 
>     * Do we still need two separate p2 update sites?

Single repo should be fine (considering the prereqs of releng being unified and regular releases)

> 
>     * maybe some more ...

Wikitext is used a lot (could be even mostly) via  Maven Central https://mvnrepository.com/artifact/org.eclipse.mylyn.wikitext - so the release process should keep the publishing to Maven Central working.

gnl42:
> > Did you copy/past the files or merge repositories? I don't see any git history for the new files
> 
> Yes because I only want to move one project.

You can still just "merge" one folder/files. Something like:  https://medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b

gnl42:
> > Could you add org.eclipse.mylyn.commons.core/.settings/{org.eclipse.jdt.core.prefs,org.eclipse.jdt.ui.prefs} to the .settings folder
> 
> I have copied the .settings folder from org.eclipse.mylyn.gitlab.core. This folder only contains the org.eclipse.jdt.core.prefs. So I have to create the org.eclipse.jdt.ui.prefs also in the gitlab projects.

With #200 I made org.eclipse.mylyn.commons.core the "reference" project for formatting rules. Not an issue at the moment, but should make a note to run the synch script occasionally

BeckerFrank:
I think we should close this pull request because I star with integrating the whole mylyn docs repository with integrating  Jenkins maven and GitHub.

ruspl-afed:
I pay tribute to your courage @BeckerFrank

wimjongman:
Where is a description of this task? It says #313, but that moves back to this PR. What is the plan, and why do we want this?

wimjongman:
Sorry, I was confused. It turns out that Github now also shows PR's when searching in the issues section. There is #131 that actually explains it.

BeckerFrank:
I close this pull request because the work was done with #316

Issue #315, opened 2023-11-21
Merge Mylyn Docs git Repository into Mylyn git Repository


BeckerFrank:
Instead of working on #313 the plan is now first to move the whole Mylyn Docs Repository into the Mylyn Repository.

Here the planed steps.

- [x] migrate the Repository and make sure that we have a working Workspace setup and a combined Jenkins Build with all combined stages
- [ ] setup Mylyn Docs only GitHub Actions in Mylyn
- [x] rework the maven definition so that we have one maven build
- [x] rework the Jenkins files so that we can perform all previous possible steps from one Jenkinsfile
- [x] rework the p2 Update Sites
- [ ] rework the html Website and GitHub pages to point to the new stuff

akurtakov:
Should mylyn.docs git repo be archived now? At least smth should be done to note that it should no longer be used.

merks:
I just remembered that the mylyn docs is using the old feature image:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/main/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext/feature.gif

rather than the new one:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/main/mylyn.builds/org.eclipse.mylyn.builds.core/mylyn.png

It will of course be good to make that consistently the new one...

merks:
It seems a little scary late to do it now...

ruspl-afed:
Do you plan to update Mylyn Docs contribution for SimRel 2023-12 @akurtakov ?

akurtakov:
@merks do we need such a release for dependencies sake? If not I don't plan to.

merks:
The existing contribution does not result in any duplicates, so I see no compelling reason to create work for yourself.

ruspl-afed:
@BeckerFrank it seems that your effort to merge Mylyn Docs into Mylyn was not brought to the attention of all the participants well enough, since we still have changes in Mylyn Docs repository. Please announce you plan more widely (perhaps, you can create a dedicated discussion item) and please ensure that we don't have to support two copies of Mylyn Docs.
Since final Mylyn 4.1 drop for SimRel 2023-12 is expected today, it may be ideal timeframe to merge Mylyn and Mylyn Docs into one project.
cc: @merks

ruspl-afed:
Mylyn Release contribution was done

ruspl-afed:
[...]

PR #316, opened 2023-11-21
#315 merge mylyn docs

BeckerFrank:
With this pull request we merge the repository mylyn docs into mylyn and setup a new Jenkins pipeline with all the mylyn doc stages and then all mylyn stages. The mylyn.setup is also changed.

This is the base for all the other tasks from #315


wimjongman (approved):
Let's go. Don't worry about the legal of that docbot user.

PR #318, opened 2023-11-23
[#315] Deploy Docs Snapshot was not executed and prepare separate Branch for mylyn.setup tests

BeckerFrank:
I will use the Branch 315-setup-test to test tat a new Eclipse Installation with Eclipse-Installer has no problems.
When finish, I will remove the branch from mylyn.setup so that we have only the main stream.


ruspl-afed (approved):
+1

PR #323, opened 2023-11-24
Remove wrong/unused/etc. maven plugin configurattions

akurtakov:


merks:
It’s really great to see this activity. Kudos! 🥇

akurtakov:
I enjoyed Mylyn a lot back in the bugzilla days. I really hope that same happens now with GH.

ruspl-afed:
Thank you for your effort @akurtakov!

This is also very helpful to learn how maven plugin configuration should change to follow Tycho upgrades.

akurtakov:
> Thank you for your effort @akurtakov!
> 
> This is also very helpful to learn how maven plugin configuration should change to follow Tycho upgrades.

One can use lemminx-maven via m2e pom.xml editor and it will warn you about non-supported configurations.

PR #324, opened 2023-11-24
Reduce ancient Orbit dependencies usage

akurtakov:
* Hamcrest 2.x is in latest platform/orbit
* javax.xml.stream is part of jvm so no need for it
* commons-logging is in latest orbit


merks:
It’s so great. I thought I was going to cry when I saw all the old dependencies.

PR #325, opened 2023-11-25
Streamline libraries usage

akurtakov:
* Move a bunch of them to latest Orbit repo
* Port from commons.lang (2.x) to commons.lang3 and no longer refer and include the former


ruspl-afed:
@akurtakov I suspect that this change was not costless for us: build became about 10 mins longer after merging this and sometime we started to have timeout failures (see https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/170/)

I tracked the problem down to this test https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/9cd5f642ad0763eba508a800a73dbd930ced10dc/mylyn.tasks/org.eclipse.mylyn.tasks.tests/src/org/eclipse/mylyn/tasks/tests/TaskDataStoreTest.java#L309 , that does some random string composition in a long cycle.

Still not sure how to deal with this better having less than one day before final contribution
cc: @merks

PR #331, opened 2023-12-01
[#315] mylyn.setup corrections

BeckerFrank:
Wit this we will reduce the number of errors in the problems view.
This is done with changes in the .project definitions and changes in mylyn.setup.

Sometimes I need to do a Projects Clean... to bring the errors to zero

So if you use the Eclipse installer and select Mylyn with the MylynDocs-Setup-Test stream, you should see fewer errors in the problem view than in the main stream.

### DIFF
--- a/mylyn.docs/.project
+++ b/mylyn.docs/.project
@@ -14,4 +14,15 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
+			<id>1701319945848</id>
+			<name></name>
+			<type>26</type>
+			<matcher>
+				<id>org.eclipse.ui.ide.multiFilter</id>
+				<arguments>1.0-name-matches-false-false-org.eclipse.mylyn*</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
 </projectDescription>
--- a/mylyn.docs/wikitext/core/.project
+++ b/mylyn.docs/wikitext/core/.project
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>core</name>
+	<name>wikitext core</name>
 	<comment></comment>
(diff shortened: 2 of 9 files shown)


ruspl-afed:
Could you please add a short description to PR to let us understand better what kind of problem you are solving @BeckerFrank

merks:
Looks promising.

gnl42:
Should the '<filteredResources>' entries really be there?

BeckerFrank:
> Should the '' entries really be there?

Where do you see the '' entries?

mylyn.docs/wikitext/.project
```diff
@@ -14,4 +14,15 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
```
akurtakov, line 17: What is the goal of this filteredResources? First time I see it so I'm just curious.

BeckerFrank: The filteredResources are used to avoid duplicate Projects in a Workspace.

Here an Example with a  project test that includes a project TestBundle1 within a folder bundle
Bildschirmfoto 2023-12-02 um 07 44 05

For example, you will receive duplicate results when searching in the workspace.

ruspl-afed:
What about my comments regarding project names with spaces?

gnl42 (approved):

BeckerFrank:
> What about my comments regarding project names with spaces?

Unfortunately, I can't find such a comment. What have I missed?

mylyn.docs/wikitext/.project
BeckerFrank: We have filterdResources in many projects to remove the target folder from the ProjectExplorer view.
I think I have found all the places where we have a grouping parent folder and add an additional filter there for folders with a .project file.

Please let me know if you find a duplicate project.

ruspl-afed (approved):
LGTM

PR #332, opened 2023-12-04
Run dependabot for mylyn.docs too

akurtakov:
Once build is merged this extra config can be removed.


akurtakov:
Merge as dependabot will not be run and PRs created from a PR.

PR #340, opened 2023-12-19
[#339] Replace dependency to org.eclipse.osgi.services

### LINKED ISSUE #339: Replace dependency to org.eclipse.osgi.services
The bundle `o.e.osgi.services` is deprecated for removal as part of https://github.com/eclipse-equinox/equinox/issues/18 and already has been changed to be empty and only re-export the 'official' OSGi artifacts published to Maven-Central.
It should be replaced by imports of the desired `org.osgi.service.*` packages so that the runtime can choose a suitable provider independently of the bundle name.

ruspl-afed:
Move to
```
Import-Package: org.osgi.service.event;version="1.4.1"
```

Fixed #339

### DIFF
--- a/mylyn.builds/org.eclipse.mylyn.builds.ui/META-INF/MANIFEST.MF
+++ b/mylyn.builds/org.eclipse.mylyn.builds.ui/META-INF/MANIFEST.MF
@@ -38,7 +38,6 @@ Require-Bundle: org.eclipse.compare;bundle-version="0.0.0",
  org.eclipse.ui.console;bundle-version="0.0.0";resolution:=optional,
  org.eclipse.ui.forms;bundle-version="0.0.0",
  org.eclipse.ui.navigator;bundle-version="0.0.0",
- org.eclipse.osgi.services;bundle-version="[3.11.100,4.0.0)",
  org.eclipse.e4.core.contexts;bundle-version="[1.12.0,2.0.0)",
  org.eclipse.e4.ui.workbench;bundle-version="[1.15.0,2.0.0)"
 Bundle-ActivationPolicy: lazy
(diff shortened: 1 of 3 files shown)


mylyn.builds/org.eclipse.mylyn.builds.ui/META-INF/MANIFEST.MF
```diff
@@ -57,4 +56,5 @@ Export-Package: org.eclipse.mylyn.builds.ui;x-internal:=true,
  org.eclipse.mylyn.internal.builds.ui.notifications;x-internal:=true,
  org.eclipse.mylyn.internal.builds.ui.util;x-internal:=true,
  org.eclipse.mylyn.internal.builds.ui.view;x-internal:=true
+Import-Package: org.osgi.service.event;version="1.4.1"
```
HannesWell, line 59: In general it is good practice to specify an upper bound to follow semantic versioning.
```suggestion
Import-Package: org.osgi.service.event;version="[1.4.0,2.0.0)"
```

HannesWell (commented):
Thanks for working on this so quickly.
The affected bundles match the analysis done by Ed using the CBI Analyser, so that looks good.
But I would specify an upper bound for the imports.

mylyn.builds/org.eclipse.mylyn.builds.ui/META-INF/MANIFEST.MF
ruspl-afed: Might be good improvement for PDE :)

HannesWell: We are working on it :)
https://github.com/eclipse-pde/eclipse.pde/pull/1007#pullrequestreview-1782918647

ruspl-afed: Great! I have improved the constraints according to your recommendations.

HannesWell (approved):
Looks good to me.
Just one tiny nit-pick, the micro version segment can usually be zero as well.

PR #344, opened 2023-12-21
[#315] releng change to one maven build task

BeckerFrank:
After we have one maven build we can look forward to do more cleanup

- one instead of two p2 updatesites
- try to reduce the .github actions / workflows
- try to convert the mylyn docs projects to build pomless


Jenkinsfile
```diff
@@ -53,38 +53,6 @@ pipeline {
 				}
 			}
 		}
-		stage('Build Docs') {
-			steps {
-				withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {
-				wrap([$class: 'Xvnc', useXauthority: true]) {
-					sh 'cd mylyn.docs && mvn clean verify -B -Psign -Dmaven.repo.local=$WORKSPACE/.m2/repository -Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true -Ddash.fail=false -Dgpg.passphrase="${KEYRING_PASSPHRASE}"'
```
ruspl-afed, line 60: And how do we build Mylyn Docs after this change @BeckerFrank ?

merks (approved):
Looks okay.

ruspl-afed:
> Looks okay.

Do you see how we build Mylyn Docs after this change @merks ? Perhaps I'm missing something

merks:
As far as I see the repositories are still separate and promoted by the `Deploy Docs Snapshot` stage.

```
Deploy Docs Snapshot
```

I am **assuming** that everything build/maven related is being done in a single invocation.

In any case, I travel in a while so I have limited capacity to look at details so when I say it looks okay, I mean that I don't see anything screaming out as wrong just looking at deltas in a browser...

ruspl-afed (changes requested):
Please make sure that we build Mylyn Docs in a single invocation of maven together with Mylyn.

Jenkinsfile
BeckerFrank: Good catch, by mistake I did not include the change in org.eclipse.mylyn/pom.xml.

ruspl-afed:
It's not as simple as it might seem, is it? This is why I had concerns @merks 

TBH I do not understand the profit from having Mylyn Docs bundles in this "special" form that requires extra steps for the rest of tooling except pure maven. Shouldn't we turn them to be "normal" bundles and simplify support a lot @akurtakov ?

merks:
There is a lot to be said for uniform structure across all the projects.  I had a really hard to wrapping my head around how to get those projects error free in a workspace.  I'm not sure what if any are the advantages of that special structure.

ruspl-afed (changes requested):
The changes goes to the right direction, but we need to repair dash check as well @BeckerFrank

BeckerFrank:
> Do you have any good reason not to commit `MANIFEST.MF` for these "special" bundles and then have a lot of things simplified @BeckerFrank ?
What do you mean with "special" bundles @ruspl-afed ?
I did not create any new bundles.

merks:
As an example, there is no MANIFEST.MF in this project:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc

All the dependencies are specified in the pom.xml:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/main/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/pom.xml

This project doesn't even have a pom.xml and expresses dependencies in the MANIFEST.MF

https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/mylyn.commons/org.eclipse.mylyn.commons.core

It looks like pom-first versus pomless.  I.e., it looks like a strange mixture...

ruspl-afed:
> What do you mean with "special" bundles @ruspl-afed ?
> I did not create any new bundles.

I mean the bundles transferred from Mylyn Docs repo. In a sense, they are "new" here.
Thanks @merks for providing an excellent problem explanation above.

My suggestion is to switch Mylyn Docs bundles from `MANIFEST.MF`less to a pomless form, exactly like you advertised in this PR description
> try to convert the mylyn docs projects to build pomless

BeckerFrank:
> As an example, there is no MANIFEST.MF in this project:
> 
> https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc
> 
> All the dependencies are specified in the pom.xml:
> 
> https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/main/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/pom.xml
> 
> This project doesn't even have a pom.xml and expresses dependencies in the MANIFEST.MF
> 
> https://github.com/eclipse-mylyn/org.eclipse.mylyn/tree/main/mylyn.commons/org.eclipse.mylyn.commons.core
> 
> It looks like pom-first versus pomless. I.e., it looks like a strange mixture...

Yes, but I would like to eliminate this in one of the next steps if possible (see the list of points in the description).

ruspl-afed:
> Yes, but I would like to eliminate this in one of the next steps if possible 

The sequence of steps is up to you of cause, but the pipeline has to be green before we can merge it.

BeckerFrank:
Sorry, I did not found the reason why org.eclipse.mylyn.docs:org.eclipse.mylyn.wikitext.asciidoc.tests:4.2.0-SNAPSHOT could not be resolver during License vetting but continuous-integration has no problems.

Any idea?

ruspl-afed:
> Any idea?

I'll have a look

akurtakov:
Having proper pom.xml files for wikitext is important as it's primary channel is(or at least used to be) Maven central and the eclipse plugin being extra on top of it.

ruspl-afed:
>> Any idea?

> I'll have a look

Now Mylyn Docs bundles have "eclipse-plugin" packaging and have MANIFEST.MF persisted, so they are a kind of normies and we can process them in a standard way.

I think it will be easier to close this PR and restart the activity with another PR @BeckerFrank 
We need to adapt (simplify) Mylyn.setup actions for `mylyn.docs` folder content and your help would be very valuable

Issue #347, opened 2024-01-05
Rework Mylyn Docs bundles from Bnd to Eclipse Plug-in form  


ruspl-afed:
After merging Mylyn Docs sources to Mylyn repository we have two bundle representations and two ways to build them. It seems very inefficient from project resources perspective to have these alternatives, because it requires much more complicated developer's setup and releng scripts than it could be in the case of unification, see #344 

If you aware of any reason to not go for unified bundle representation and releng manifests please let me know @merks @akurtakov 
With the current state of things we have a risk to not provide actual Mylyn Docs contribution in time for SimRel 2024-03 M1

akurtakov:
Wikitext non-ui parts has this way of building as it used to have bigger usage outside of Eclipse IDE  via Maven Central thus having complete pom.xml files is crucial. Whether it's done via maven-bundle-plugin or via some other plugin (tycho, bnd, handcrafted manifests, etc.) is another question but whatever path is taken it would require significant overhaul to get things working. 
I had quite a rough path getting it to current (sad) state (last big effort was https://github.com/eclipse-mylyn/org.eclipse.mylyn.docs/commit/ef150683f283e58e8e57055d552189d73ed3b085 ) but it was even worse.
Mylyn Docs is already built  and published as separate Maven call but from same Jenkins so why is this a blocker for M1?

ruspl-afed:
Thank you for letting me know about Maven Central publishing @akurtakov , now I understand that we need to keep pom.xml for each bundle.

> Mylyn Docs is already built and published as separate Maven call but from same Jenkins so why is this a blocker for M1?

There is an intention in #344 to simplify build scripts and have only one maven call. The intention itself looks rational to me, but I blocked the merge of mentioned PR to not break the working state of things. Since it may be not obvious for other committers how this or similar change can break the M1 contribution - I can say we are at risk.

At the current time point I can suggest to delay both #344 and this one and contribute Mylyn Docs to from master as it is now. Then we can have a bit more time to resolve this one, may be till M3.

ruspl-afed:
> Mylyn Docs is already built and published as separate Maven call but from same Jenkins so why is this a blocker for M1?

Yes, for M1 contributing snapshot may be fine, I did it with https://github.com/eclipse-simrel/simrel.build/pull/157. But for Release contribution we currently don't have any actual configuration.

PR #363, opened 2024-01-11
[#347] adjust the OOMPH setup

BeckerFrank:
To continue with  issue #315 we have to adjust the Mylyn.setup to the changes from #347


ruspl-afed:
But do we need these launches? Mylyn Docs bundles now have MANIFEST.MF

BeckerFrank:
> But do we need these launches? Mylyn Docs bundles now have MANIFEST.MF

Yes we have the help project and the launcher "mylyn verify" creates the missing files.

ruspl-afed:
Could you please point me on it? I'm asking to ensure that nothing was missed during conversion of maven packaging.

ruspl-afed:
In other words: is this necessary for installer to call the complete maven build for the whole repository?

ruspl-afed:
Thank you for the explanation, now I see. 
So we need to find a way how to build this only module instead of running the whole build.

ruspl-afed (approved):
Since it is checked on your side I think it is good to go.
However, reducing the scope of initial build could be a significant improvement.

BeckerFrank:
> Since it is checked on your side I think it is good to go. However, reducing the scope of initial build could be a significant improvement.

Yes, I can try if I find a better way starting mid of next week.

PR #365, opened 2024-01-13
OOMPH setup: cleanup initial mvn build

BeckerFrank:


ruspl-afed:
Good to see the scope of maven build reduced, thank you!

I still have concerns why building with maven is necessary prior to import of projects after having MANIFEST.MF in place. 
If some resources are not present, may be it can be fixed with adding m2e build for help bundle. What kind of errors do we have without maven build?

BeckerFrank:
> Good to see the scope of maven build reduced, thank you!
> 
> I still have concerns why building with maven is necessary prior to import of projects after having MANIFEST.MF in place. If some resources are not present, may be it can be fixed with adding m2e build for help bundle. What kind of errors do we have without maven build?

The problem is that the following two files are created by the maven build 
```
		<toc file="help/Mylyn-WikiText-User-Guide-toc.xml" primary="false"/>
		<toc file="help/devguide/WikiText-Developer-Guide-toc.xml" primary="false"/>
```

please look into the maven-antrun-plugin execution in the pom.xml lines 59 - 68

merks:
I'm curious if it might be okay to create empty or place-holders files for these?  It's possible to have tasks to create resources directly in the clone.  For example, for the SimRel setup, I don't want to check in a bunch of files in the .github repository clone so I create the .project (so the clone can be imported into the workspace as a project), .settings/org.eclipse.core.resources.prefs (so that the project-specific encoding warning is gone), and .gitignore (to ignore those two resources and the .gitignore itself):

https://github.com/eclipse-simrel/simrel.build/blob/99f6ab93f8b02a8757039baf66d423f720c3fc08/SimRel.setup#L219-L254

ruspl-afed:
Following Ed's comment, 
do you think we can "echo" these 2 files with minimal valid content from setup script, to then have them fulfilled by M2E builder from IDE @BeckerFrank ?

BeckerFrank:
> Following Ed's comment, do you think we can "echo" these 2 files with minimal valid content from setup script, to then have them fulfilled by M2E builder from IDE @BeckerFrank ?

Yes, I'll try, but I won't be able to do it for the next two days as I'm on a business trip. But maybe I'll be able to do it before then.

ruspl-afed:
We have no hurry here, this is just nice to have. Have a productive business trip.

Issue #386, opened 2024-01-30
Clean Up: Resynch templates


gnl42:
Synchronize the code cleanup templates

PR #387, opened 2024-01-30
386: Clean Up: Resynch templates

gnl42:
Minor change to the 'New Class' template

Resynch the 'Clean up...' and 'Formatter' templates for all the java projects

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/386


ruspl-afed:
Thank you for your effort @gnl42 
But I'm not sure how a human being would be able to verify such a difference.

gnl42:
@ruspl-afed commons.core is the master source. You just need to look at the files in its .settings folder. 
Those files are copied into all the objects if they have changed from the master. 

One of the main motivation for this was to synch the new docs folders

ruspl-afed:
GitHub UI is not helpful to understand the diff for `org.eclipse.jdt.ui.text.custom_code_templates` property value. I was forced to extract it to a file, format and compare.
And I can see only one diff there:
image

ruspl-afed (approved):
LGTM

gnl42:
@ruspl-afed The real impact are the wikitext projects since they were brought in after the first sync

ruspl-afed:
Yes, I saw it. But that's fine, because no one but us will take care of them.

Issue #375, opened 2024-01-30
Clean Up: Commons


gnl42:
Apply code cleanup templates

PR #388, opened 2024-01-30
Clean Up: Commons #375

gnl42:
Apply Clean Up rules to Commons packages

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/375


gnl42:
Hmm, came across some unwanted "static final" transformations. 
Clean up is cleaning up too much

akurtakov:
What I can see is "public static final" being removed in interfaces as these are implicit which is OK IMHO.

akurtakov (approved):

ruspl-afed (approved):
I'm not always happy with combined catch block , but for this code it looks appropriate
LGTM

gnl42:
> What I can see is "public static final" being removed in interfaces as these are implicit which is OK IMHO.

Good catch. Forgot about that, I don't like having constants in interfaces so it's been a long time.

Issue #389, opened 2024-01-31
Explicit lucene version in org.eclipse.mylyn.tasks.index.core


gnl42:
Not sure why doing a Clean Up on Tasks made this one pop up

org.eclipse.mylyn.tasks.index.core is asking for lucene 9.5.0 (don't know why). What we get from p2 is 9.9.1

ruspl-afed:
> asking for lucene 9.5.0 (don't know why)

it is asking for it to not get older versions from platform, that was a case some time ago. If you think this is an issue, feel free to create a PR

ruspl-afed:
Interesting, perhaps there was a change of exported packages on lucene side or something similar.

> Changing the version in the manifest to 0.0.0 gets rid of the error.

Could you please create a PR with this very change to let us see what pipeline tests think about it?

merks:
The 9.5.0 is just a lower bound.  I built new version of lucene in the nightly build yesterday:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/18e022db-64fb-4016-8f1b-40940d38ffd1)

I wonder which version is in your target platform?

It's strange that removing the bound would have any impact on the error.  Does that change which version is show in the plug-in's dependencies?

gnl42:
No, in either case its using 9.9.1

ruspl-afed:
> The 9.5.0 is just a lower bound.

Yes, true. It was an actual/ version of Apache Lucene when we (probably, I) did a reanimation procedures for Mylyn.

merks:
Very strange.  The problem you show is generally fixed by importing the package of the "indirectly required" class in the MANIFEST.MF...

ruspl-afed:
May it be the consequence of clean-up, that cleared some declarations, then imports, and finally changed the resolution chain on IDE side?

merks:
Maybe there is a split package involved because I see that org.apache.lucene.store is already imported.  Let me look...

merks:
This just makes no sense to me.  If I pull the latest, should I be able to reproduce this?

gnl42:
Clean up only affects the java classes. maven build has the same problem

gnl42:
@merks Yes. Just select the Tasks node, right click, and select Cleanup

merks:
I end up with a rather large set of errors:

[...]

PR #390, opened 2024-01-31
Explicit lucene version #389

gnl42:
Strange missing class error after doing a Clean Up of Tasks

Tasks is using an explicit reference to 9.5.0, p2 -> 9.5.1.
Changing the version to 0.0 .0 clears the error

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/389

### DIFF
--- a/mylyn.tasks/org.eclipse.mylyn.tasks.index.core/META-INF/MANIFEST.MF
+++ b/mylyn.tasks/org.eclipse.mylyn.tasks.index.core/META-INF/MANIFEST.MF
@@ -10,14 +10,14 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-ActivationPolicy: lazy
 Export-Package: org.eclipse.mylyn.internal.tasks.index.core;x-internal:=true
-Import-Package: org.apache.lucene.analysis;version="9.5.0",
- org.apache.lucene.analysis.core;version="9.5.0",
- org.apache.lucene.analysis.miscellaneous;version="9.5.0",
- org.apache.lucene.analysis.standard;version="9.5.0",
- org.apache.lucene.document;version="9.5.0",
- org.apache.lucene.index;version="9.5.0",
- org.apache.lucene.queryparser.classic;version="9.5.0",
- org.apache.lucene.search;version="9.5.0",
- org.apache.lucene.store;version="9.5.0",
- org.apache.lucene.util;version="9.5.0"
+Import-Package: org.apache.lucene.analysis;version="0.0.0",
+ org.apache.lucene.analysis.core;version="0.0.0",
+ org.apache.lucene.analysis.miscellaneous;version="0.0.0",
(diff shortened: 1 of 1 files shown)


ruspl-afed (approved):
I don't have strong objections.
WDYT @merks ?

merks:
Give me a few minutes.  I'd like to understand how this is possible.

gnl42:
It's possible you may have clean up some bad conversions

merks:
I don't know which project had the error being fixed:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/389#issuecomment-1918636196

I don't have a big objection to doing what's necessary to fix the errors, but it all seems very strange to me.

akurtakov:
Not having lower boundary opens the door for a case where the package is there but the API is not what we expect. I understand this is quite unlikely as Platform will bring new enough Lucene in most products but still something to be considered.

mylyn.tasks/org.eclipse.mylyn.tasks.index.core/META-INF/MANIFEST.MF
```diff
@@ -10,14 +10,14 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
...
- org.apache.lucene.analysis.core;version="9.5.0",
- org.apache.lucene.analysis.miscellaneous;version="9.5.0",
- org.apache.lucene.analysis.standard;version="9.5.0",
- org.apache.lucene.document;version="9.5.0",
- org.apache.lucene.index;version="9.5.0",
- org.apache.lucene.queryparser.classic;version="9.5.0",
- org.apache.lucene.search;version="9.5.0",
- org.apache.lucene.store;version="9.5.0",
- org.apache.lucene.util;version="9.5.0"
+Import-Package: org.apache.lucene.analysis;version="0.0.0",
```
laeubi, line 13: You should not use a version of 0.0.0 here, thats the default and only highly confusing.

laeubi:
I don't understand what this should fix either... 9.5.0 is a open range and includes every higher version (including 9.5.1), this fix looks suspicious and more likely is a sign of wrong dependency chains.

mylyn.tasks/org.eclipse.mylyn.tasks.index.core/META-INF/MANIFEST.MF
merks: Unfortunately the discussion is split better the PR and the issue.  I think the appropriate fix is what I show here:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/389#issuecomment-1918636196

ruspl-afed (commented):
@gnl42 what if try to import Lucene package(s) for bundles that started to require it after clean-up?

gnl42:
Adding explicit import to the test project seems to fix the issue. Somthing to watch out for

laeubi:
> Adding explicit import to the test project seems to fix the issue. Somthing to watch out for

Nooo please see my comment please check if you need to update the BREE instead!

Issue #376, opened 2024-01-30
Clean Up: Tasks


gnl42:
Apply code cleanup templates

PR #391, opened 2024-01-31
Clean up: Tasks #376

gnl42:
Clean Up: Tasks

Apply cleanup templates to the Tasks modules

Also found some clean up bugs that I've collected and will pass on to the clean up crew


akurtakov:
I understand it's easy to run once per module but reviewing 600+ files is just not feasible. As these are generated ones feel free to push them yourself if you're confident but you can't expect anyone to review such PRs - you would have to go to a far more grainer level (patch size) for that.

merks:
Yes, it's too much to review.  If the build is good and you've glanced at the changes yourself, we can assume it's okay.

gnl42:
Thanks. I mainly just wanted let you now what was going on. And maybe get some feedback on the cleanup results

PR #429, opened 2024-02-06
419 resolve light warnings for mylyn reviews

gnl42:


gnl42:
@BeckerFrank Is it possible to rerun the fail tasks? Or do I have push a fake change to re-trigger them?

PR #436, opened 2024-02-16
add a gitlab junit project with first 6 tests

BeckerFrank:
All test are ignored when we can not found a Gitlab Test instance.


mylyn.tasks/connectors/gitlab/org.eclipse.mylyn.gitlab.core.test/src/org/eclipse/mylyn/gitlab/core/test/GitlabTestFixture.java
```diff
@@ -0,0 +1,126 @@
+/*******************************************************************************
+ * Copyright © 2024 frank
```
ruspl-afed, line 2: please add more details

mylyn.tasks/connectors/gitlab/org.eclipse.mylyn.gitlab.core.test/build.properties
```diff
@@ -0,0 +1,4 @@
+source.. = src/
```
ruspl-afed, line 1: please add header

ruspl-afed (commented):
Thank you for working on this @BeckerFrank !

I have only minor comments regarding copyright headers.

mylyn.tasks/connectors/gitlab/org.eclipse.mylyn.gitlab.core.test/src/org/eclipse/mylyn/gitlab/core/test/GitlabTestFixture.java
BeckerFrank: I'll change the headers.
Oddly enough, I found another place where I had a © instead of "(c)", so I'll change that too.

gnl42: My fault. I'll update the 'New Class' template

Issue #37, opened 2023-01-06
Gitlab connector


wimjongman:
We need to connect GitLab.

See https://github.com/pweingardt/mylyn-gitlab/issues/61

bgranier:
Hi,
To anwser to the question asked here : https://github.com/pweingardt/mylyn-gitlab/issues/61#issuecomment-1373426635 .

Unfortunately, not. I am teaching and this connector is "just" a way to show to students how to manage issues. So I can not do a donation.

Best regards,

Bernard

mjmeijer:
@bgranier question was not to make monetary donation, but @wimjongman asked would you donate this `code` to the revived mylyn project as an example connector? JIRA and Trac seem a bit out of date nowadays.

bgranier:
humm you mean "to give me" the code in order to revive it to get a runnable plugin ?

Why not, but i am not sure that i will have the time for that.

BeckerFrank:
> humm you mean "to give me" the code in order to revive it to get a runnable plugin ?
> 
> Why not, but i am not sure that i will have the time for that.

No we need to ask the maintainer of the Repository (https://github.com/pweingardt) if we get the permission to include a copy of the source into the new eclipse.mylyn GitHub organisation.

gnl42:
I think we can close this one

BeckerFrank:
> I think we can close this one

I haven't received any feedback from [pweingardt] (https://github.com/pweingardt), so I think we should create a new mylyn Task Connector for Gitlab.

I have created the not officially supported Bugzilla REST Connector, so I know how to implement a Task Connector.

Should I start developing an MVP version?

wimjongman:
Sorry for the confusion. Let me try to explain my request a bit better.

It looks like you have built a connector for GL. We think a GL connector is important.

Indeed, my request was not for money (thanks Maarten), but for your code. I should have called it code contribution and not donation. Sorry for the confusion.

If you want, you can become an Eclipse Mylyn committer and help maintain the connector.

We did not look at your code yet, so I am not sure if it is even possible.

Some of the things that should happen are:

* You have to license the code under EPL
* Your code has to move to the org.eclipse namespace.
* You have to sign the eclipse contributor agreement and declare that all the copyright of the code belongs to you and that you contribute the code to the open source.

There are probably a couple of other things that I have forgotten but these are the basics.

BeckerFrank:
[...]

PR #461, opened 2024-03-02
[#37] run clean up... for all Gitlab projects

BeckerFrank:


mylyn.tasks/connectors/gitlab/org.eclipse.mylyn.gitlab.core/src/org/eclipse/mylyn/internal/gitlab/core/GitlabRepositoryConnector.java
```diff
@@ -119,18 +118,9 @@ private boolean ignoredProperty(String propertyName) {
 		return false;
 	}
 
-	private final PropertyChangeListener repositoryChangeListener4ConfigurationCache = new PropertyChangeListener() {
+	private PropertyChangeListener repositoryChangeListener4ConfigurationCache = null;
```
ruspl-afed, line 121: Is this really a result of clean up? It is strange to see it mutable and nullable instead of the recent final non-nullable.

BeckerFrank: Yes, if look strange but  the Clean Up generates an error.

Cannot reference a field before it is defined. in File GitlabRepositoryConnector.java line 315.

this.clientCache.invalidate(new RepositoryKey(taskRepository));
is changed to
clientCache.invalidate(new RepositoryKey(taskRepository));

ruspl-afed: This Clean Up implementation looks weird: making field mutable cannot be claimed as "Clean Up".

ruspl-afed: However, if you can make them `final` back, this one is good to go.

BeckerFrank: > However, if you can make them `final` back, this one is good to go.

No that is not possible we need to have repositoryChangeListener4ConfigurationCache and repositoryChangeListener4ClientCache none final because we otherwise get an  error

`The blank final field ... may not have been initialized.`

We have configured the Save Actions to do a Clean up and we can not make any changes in file  GitlabRepositoryConnector.java because of the error in (remove `this.`)

```
	private final PropertyChangeListener repositoryChangeListener4ClientCache = evt -> {
		TaskRepository taskRepository = (TaskRepository) evt.getSource();
		clientCache.invalidate(new RepositoryKey(taskRepository));
	};
```

merks: Why would one explicitly initialize to null when that is implicit?  If you check in a debugger, that actually generates code to assign the null to field that is already null.

BeckerFrank: > Why would one explicitly initialize to null when that is implicit? If you check in a debugger, that actually generates code to assign the null to field that is already null.

Good catch, I just ran the unit tests and could not find this error.
Next commit will fix this.

ruspl-afed:
See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4382

ruspl-afed (approved):
Good to go at this stage

Issue #459, opened 2024-03-02
Restore releng for publishing Maven Docs to maven central


ruspl-afed:
Maven publishing is most likely not working after moving Mylyn Docs inside Maven and further releng work. Perhaps, we need to restore it.

@akurtakov how do you recommend to approach this? Should we have a separate Jenkins config that launches `mvn` from the specific entry point (special pom.xml with enumerated modules to publish) and with special set of options?
An outline of the steps to implement would be very helpful, because I'm not an experienced maven publisher.

akurtakov:
This was fully automated via https://github.com/eclipse-mylyn/org.eclipse.mylyn.docs/blob/master/Jenkinsfile-release so running some job using this jenkinsfile should be a good starting point.

BeckerFrank:
How can this work?

We now have version 3.0.48.202308291007 as a dependency. But build the version 4.3 and include this version in the updatesite!

ruspl-afed:
Maven Central and Eclipse Update Site are totally different things. Update site is easy to produce and committer rights are sufficient to upload. While uploading to Maven Central is quite a quest.

BeckerFrank:
> Maven Central and Eclipse Update Site are totally different things. Update site is easy to produce and committer rights are sufficient to upload. While uploading to Maven Central is quite a quest.

That is not the point.
We use  `mvn clean verify` for the Mylyn build, but Mylyn.docs needs `mvn clean deploy` to push the build result into the remote repository.

Why do you fix set the version to the old version (3.0.48.202308291007) for org.eclipse.mylyn.wikitext and org.eclipse.mylyn.wikitext.textile in the pom of org.eclipse.mylyn.wikitext.maven?

How are newer version of  the projects org.eclipse.mylyn.wikitext and org.eclipse.mylyn.wikitext.textile build and come into the maven repository?

ruspl-afed:
Perhaps I miss something important: do you mean that you have a solution how to restore Mylyn Docs publishing? If so, please provide a PR.

ruspl-afed:
@BeckerFrank  I created a PR #487 to revert versions back to available in reactor

BeckerFrank:
> Perhaps I miss something important: do you mean that you have a solution how to restore Mylyn Docs publishing? If so, please provide a PR.

We have the following calls to maven for builder Mylyn

```
mvn \
clean \
verify \
-B \
$MAVEN_PROFILES \
[...]

PR #487, opened 2024-04-07
[#459] Restore releng for publishing Maven Docs to maven central

ruspl-afed:
Revert the fix for wikitext maven plugin dependencies


ruspl-afed:
@BeckerFrank I'm merging this to M1

Issue #492, opened 2024-04-18
Task search index is broken after platform update, attempts to load a non-existing lucene version.


gvl2023:
### Discussed in https://github.com/eclipse-mylyn/org.eclipse.mylyn/discussions/479

<div type='discussions-op-text'>

<sup>Originally posted by **gvl2023** April  1, 2024</sup>
Getting the following error after updating to the 4.31 simultaneous release. Happens on startup and task list synchronization. More of a  nuisance as it doesn't appear to affect basic functionalities. Any workarounds?

```
An internal error occurred during: "Task List Indexer".

java.lang.IllegalArgumentException: Could not load codec 'Lucene95'. Did you forget to add lucene-backward-codecs.jar?
	at org.apache.lucene.index.SegmentInfos.readCodec(SegmentInfos.java:520)
	at org.apache.lucene.index.SegmentInfos.parseSegmentInfos(SegmentInfos.java:409)
	at org.apache.lucene.index.SegmentInfos.readCommit(SegmentInfos.java:368)
	at org.apache.lucene.index.SegmentInfos.readCommit(SegmentInfos.java:304)
	at org.apache.lucene.index.SegmentInfos$1.doBody(SegmentInfos.java:546)
	at org.apache.lucene.index.SegmentInfos$1.doBody(SegmentInfos.java:543)
	at org.apache.lucene.index.SegmentInfos$FindSegmentsFile.run(SegmentInfos.java:820)
	at org.apache.lucene.index.SegmentInfos$FindSegmentsFile.run(SegmentInfos.java:770)
	at org.apache.lucene.index.SegmentInfos.readLatestCommit(SegmentInfos.java:548)
	at org.apache.lucene.index.SegmentInfos.readLatestCommit(SegmentInfos.java:532)
	at org.apache.lucene.index.IndexWriter.<init>(IndexWriter.java:1011)
	at org.eclipse.mylyn.internal.tasks.index.core.TaskListIndex.createIndexWriter(TaskListIndex.java:1282)
	at org.eclipse.mylyn.internal.tasks.index.core.TaskListIndex.rebuildIndexCompletely(TaskListIndex.java:1228)
	at org.eclipse.mylyn.internal.tasks.index.core.TaskListIndex.maintainIndex(TaskListIndex.java:1095)
	at org.eclipse.mylyn.internal.tasks.index.core.TaskListIndex$MaintainIndexJob.run(TaskListIndex.java:184)
[...]

Issue #464, opened 2024-03-04
An internal error occurred during: "Task List Indexer".


mjmeijer:
Constant errors in Task List Indexer:

An internal error occurred during: "Task List Indexer".
Could not load codec 'Lucene95'. Did you forget to add lucene-backward-codecs.jar?

-- Error Details --
Date: Mon Mar 04 12:42:25 CET 2024
Message: An internal error occurred during: "Task List Indexer".
Severity: Error
Product: Eclipse IDE 4.30.0.20231201-1200 (org.eclipse.epp.package.rcp.product)
Plugin: org.eclipse.core.jobs
Session Data:
eclipse.buildId=4.30.0.20231201-1200
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_NL
Framework arguments:  -product org.eclipse.epp.package.rcp.product -keyring /Users//.eclipse_keyring
Command-line arguments:  -os macosx -ws cocoa -arch x86_64 -product org.eclipse.epp.package.rcp.product -keyring /Users//.eclipse_keyring

Exception Stack Trace:
java.lang.IllegalArgumentException: Could not load codec 'Lucene95'. Did you forget to add lucene-backward-codecs.jar?
	at org.apache.lucene.index.SegmentInfos.readCodec(SegmentInfos.java:520)
	at org.apache.lucene.index.SegmentInfos.parseSegmentInfos(SegmentInfos.java:409)
	at org.apache.lucene.index.SegmentInfos.readCommit(SegmentInfos.java:368)
	at org.apache.lucene.index.SegmentInfos.readCommit(SegmentInfos.java:304)
	at org.apache.lucene.index.SegmentInfos$1.doBody(SegmentInfos.java:546)
	at org.apache.lucene.index.SegmentInfos$1.doBody(SegmentInfos.java:543)
	at org.apache.lucene.index.SegmentInfos$FindSegmentsFile.run(SegmentInfos.java:820)
	at org.apache.lucene.index.SegmentInfos$FindSegmentsFile.run(SegmentInfos.java:770)
	at org.apache.lucene.index.SegmentInfos.readLatestCommit(SegmentInfos.java:548)
	at org.apache.lucene.index.SegmentInfos.readLatestCommit(SegmentInfos.java:532)
	at org.apache.lucene.index.IndexWriter.<init>(IndexWriter.java:1011)
[...]

PR #496, opened 2024-04-25
Rebuild Task search index if needed. #492

gnl42:
Just rebuild the Lucene index if an existing one can't be read.

There is no guarantee that an old codec will be available with the lucene-backward-codecs so why bother trying

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/492


ruspl-afed (approved):
LGTM

merks (approved):
That seems fine.

Wrapping your own bundles ( the part deleted from the target platform) is  really not okay and its good it's being deleted.  If you need that, do it in Orbit.

ruspl-afed:
> Wrapping your own bundles

It was my fault. I had a vain hope that this will help us to fix [the issue](https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/464) . Will be good if we remove this.

PR #499, opened 2024-04-28
Resolve "light" warnings for Mylyn Gitlab Connector: second pass #416

gnl42:
Second pass

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/416


gnl42:
Will re-create with bigger scope

Issue #500, opened 2024-04-28
Resolve "light" warnings for Mylyn


gnl42:
Second pass at cleaning up the "light" warnings.

PR #503, opened 2024-04-30
resolve light warnings for mylyn #500

gnl42:
Catch up on the few things missing by the specific 'light warnings' issues


BeckerFrank (approved):
I have not gone through all 263 files in detail, but I have looked at some of the changes in each commit and think this is fine.

Issue #517, opened 2024-05-07
org.apache.tika.mime is missing 


merks:
My Oomph builds are failing like this:
```
[ERROR] Cannot resolve target definition:
[ERROR]   Software being installed: org.eclipse.mylyn.builds.feature.feature.group 4.3.0.v20240429-0805
[ERROR]   Missing requirement: org.eclipse.mylyn.tasks.ui 4.3.0.v20240506-2240 requires 'java.package; org.apache.tika.mime [1.24.0,2.0.0)' but it could not be found
[ERROR]   Cannot satisfy dependency: org.eclipse.mylyn.builds.feature.feature.group 4.3.0.v20240429-0805 depends on: org.eclipse.equinox.p2.iu; org.eclipse.mylyn.builds.ui 
```
Updating the Oomph target platform fails like this:
```
  ERROR: org.eclipse.equinox.p2.director code=0 Software being installed: artificial_root 1.0.0.v1715069846493
  ERROR: org.eclipse.equinox.p2.director code=0 Missing requirement: Mylyn Tasks UI 4.3.0.v20240506-2240 (org.eclipse.mylyn.tasks.ui 4.3.0.v20240506-2240) requires 'java.package; org.apache.tika.mime [1.24.0,2.0.0)' but it could not be found
  ERROR: org.eclipse.equinox.p2.director code=1 Cannot satisfy dependency:
    ERROR: org.eclipse.equinox.p2.director code=0 From: artificial_root 1.0.0.v1715069846493
    ERROR: org.eclipse.equinox.p2.director code=0 To: org.eclipse.equinox.p2.iu; org.eclipse.oomph.setup.mylyn 0.0.0
  ERROR: org.eclipse.equinox.p2.director code=1 Cannot satisfy dependency:
    ERROR: org.eclipse.equinox.p2.director code=0 From: Mylyn Build Framework 4.3.0.v20240406-0935 (org.eclipse.mylyn.builds.ui 4.3.0.v20240406-0935)
    ERROR: org.eclipse.equinox.p2.director code=0 To: osgi.bundle; org.eclipse.mylyn.tasks.ui 4.3.0
  ERROR: org.eclipse.equinox.p2.director code=1 Cannot satisfy dependency:
    ERROR: org.eclipse.equinox.p2.director code=0 From: Oomph Setup Mylyn 1.19.0.v20240507-101722 (org.eclipse.oomph.setup.mylyn 1.19.0.v20240507-101722)
[...]

PR #518, opened 2024-05-08
Add org.apache.tika.core to the category.xml

merks:
With the addition of dependencies on org.apache.tika.core, it's needed in the update site.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/517


ruspl-afed (approved):
Thank you @merks ! 
I missed this during review.

Issue #305, opened 2023-11-02
Remove guava TypeToken


gnl42:
#287 left com.google.common.reflect.TypeToken behind in GerritClient212.

Bugzilla uses com.google.gson.reflect.TypeToken for similar purposes. 

Gerrit unit tests are currently disabled so unable to confirm if the two are functionally equivalent

gnl42:
@BeckerFrank I saw you used the gson version in your GitLab code, any thoughts on this one?

BeckerFrank:
> @BeckerFrank I saw you used the gson version in your GitLab code, any thoughts on this one?

Yes, we use Gson and thus com.google.common.reflect.TypeToken to convert the REST responses from Bugzilla Rest and Gitlab Connector. This is also used in Common SDK for the test configuration.

If we should no longer use Gson what should we use for serialization and deseralisation?

gnl42:
I meant any thoughts on this issue.

Can I replace the guava TypeToken with the gson TypeToken in GerritClient212?

It's the remaining guava use Mylyn

BeckerFrank:
> I meant any thoughts on this issue.
> 
> Can I replace the guava TypeToken with the gson TypeToken in GerritClient212?
> 

Yes, that should work

PR #522, opened 2024-05-10
Remove guava TypeToken #305

gnl42:
Replace guava TypeToken with gson TypeToken

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/305


ruspl-afed (approved):
LGTM

Issue #444, opened 2024-02-21
Reduce grade of Mylyn Docs API from "public" to "internal"


ruspl-afed:
Since Mylyn Docs has major version change during this release cycle we have an unique opportunity to reduce support cost: mark all Mylyn Docs public API is internal without changing.

If anyone will complain regarding this we may consider particular signatures to be public again, otherwise they will published only after dedicated review.

ruspl-afed:
Fixed with #446 , additional warnings should be fixed with #423 or subsequent issues

PR #523, opened 2024-05-10
Don't mark wikitext API used by LSP4E as internal

akurtakov:
In versions 3.x this was api and internals are well hidden (esp in markdown plugin) thus the change to make everything internal was excessive and generates "Discouraged access" issues in LSP4E.


akurtakov:
It was identified that lsp4e is not using latest wikitext - https://github.com/eclipse-cdt/cdt/pull/779 . I tried to do that but it introduced extra discouraged access warnings thus this PR. I haven't searched why everything went internal at some point thus I'm making public only the parts that lsp4e needs that I am sure should be proper API.

ruspl-afed:
@akurtakov  don't you think this discussion in the form of a monologue was too short to promote several packages to API?

More API requires more effort from severely understaffed Mylyn team, and this is why all the Mylyn WikiText API was degraded according to https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/444

"Discouraged access" issues in LSP4E project or any other project is not a sufficient reason to promote anything to the API in Mylyn project.

With all my respect to your contributions @akurtakov my current plan is to revert this PR before 2024-06 M3, because this is not how API should be promoted, even if some package was an API in the previous major version.

cc: @merks

akurtakov:
I did it that way as it's actually partial revert of removal of API that has been such for years. As we obviously disagree I would prefer to hear from other committers (vote maybe?) before we come to conclusion.

akurtakov:
I have created https://github.com/eclipse-mylyn/org.eclipse.mylyn/discussions/524 and let the community decide :).

Issue #491, opened 2024-04-18
Unable to Add GitHub Gist as Task Repository


emecas:
### Versions:

1) Eclipse IDE for Enterprise Java and Web Developers (includes Incubating components)
Version: 2024-03 (4.31.0)
Build id: 20240307-1437

2) Eclipse IDE for Enterprise Java and Web Developers	4.31.0.20240307-1200	epp.package.jee	Eclipse Packaging Project

3) Mylyn Tasks Connector: GitHub	6.4.0.v20240209-1319	org.eclipse.mylyn.github.feature.feature.group	Eclipse Mylyn

### Problem : 

I'm encountering an error when attempting to add a new task repository of type "GitHub Gist." Regardless of the URL format used, I receive the message: `Enter a valid server URL.`

Here are the specific URLs I've tried (replacing <my-github-user> with my actual username):

    https://gist.github.com (Base URL)
    https://gist.github.com/<my-github-user> (My User Gists)
    https://gist.github.com/<other-valid-github-user> (Another User's Gists)
    https://gist.github.com/discover (Discover Gists)

Additionally, I have attempted using:

    Valid user credentials (username and password)
    Valid access token

Unfortunately, using any of these options prevents the expected functionality. The buttons "Validate Settings" and "Finish" remain disabled.

### How to reproduce:

1. Go to the contextual menu in the `Task List` tab, then click `New` and then `Query...`

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/849502/f9cffcaa-981f-45b5-ad2d-b6ed8b66ac50)

2. in the emergent dialog, click in `Add Task Repository` button, and then select from the list `GitHub Gist` and then click the`Next` button

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/849502/ed1c3678-62d7-4d8b-8b49-ef15e38ba836)

3. Here is the problem

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/849502/f456e2b7-3fcf-419a-a050-412b92788ddd)

BeckerFrank:
This is not easy to fix. We have taken over responsibility from another project. As far as I can see, the problem is that the gist URL is validated with org.eclipse.egit.github.core.RepositoryId.createFromUrl. Gist URLs have no owner and no name segment like repositories. I don't know how this has ever worked.

emecas:
[...]

PR #525, opened 2024-05-11
[#491] fix "GitHub Gist" URL Validation

BeckerFrank:


mylyn.github/org.eclipse.mylyn.github.ui/META-INF/MANIFEST.MF
```diff
@@ -30,7 +30,8 @@ Export-Package: org.eclipse.mylyn.internal.github.ui;version="6.2.1";x-internal:
  org.eclipse.mylyn.internal.github.ui.issue;version="6.2.1";x-internal:=true,
  org.eclipse.mylyn.internal.github.ui.pr;version="6.2.1";x-internal:=true
 Bundle-ActivationPolicy: lazy
-Import-Package: org.apache.commons.lang3;version="3.13.0",
+Import-Package: org.apache.commons.httpclient;version="[3.1.0,4.0.0]",
```
gnl42, line 33: Why is httpclient added?

BeckerFrank: Good catch! This can be removed. Left over from another way to fix this problem.

merks: Thank goodness because that thing is really ancient:

https://repo1.maven.org/maven2/commons-httpclient/commons-httpclient/3.1/

Issue #370, opened 2024-01-18
Make `FocusedJavaAllProposalComputer` more robust


ruspl-afed:
During attempt to complete invalid Java code there is a Mylyn failure reported:

`The 'org.eclipse.mylyn.java.ui.javaAllCompletionProposalComputer' proposal computer from the 'org.eclipse.mylyn.java.ui' plug-in did not complete normally. The extension has thrown a runtime exception.`

Perhaps this is pure JDT problem, but we can try to be more robust on Mylyn side
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.impl.Constant.typeID()" because "this.constant" is null
	at org.eclipse.jdt.internal.compiler.ast.Expression.computeConversion(Expression.java:707)
	at org.eclipse.jdt.internal.compiler.ast.SingleNameReference.computeConversion(SingleNameReference.java:320)
	at org.eclipse.jdt.internal.compiler.ast.BinaryExpression.resolveType(BinaryExpression.java:1921)
	at org.eclipse.jdt.internal.compiler.ast.CombinedBinaryExpression.resolveType(CombinedBinaryExpression.java:398)
	at org.eclipse.jdt.internal.compiler.ast.MemberValuePair.resolveTypeExpecting(MemberValuePair.java:77)
	at org.eclipse.jdt.internal.compiler.ast.Annotation.resolveType(Annotation.java:956)
	at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveAnnotations(ASTNode.java:970)
	at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveAnnotations(ASTNode.java:800)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.getAnnotationTagBits(SourceTypeBinding.java:1642)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.updateSupertypesWithAnnotations(TypeDeclaration.java:1886)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.integrateAnnotationsInHierarchy(CompilationUnitScope.java:402)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment$CompleteTypeBindingsSteps.perform(LookupEnvironment.java:173)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.completeTypeBindings(LookupEnvironment.java:605)
[...]

PR #526, opened 2024-05-12
[#370] catch the exception

BeckerFrank:


mylyn.context/org.eclipse.mylyn.java.ui/src/org/eclipse/mylyn/internal/java/ui/editor/FocusedJavaAllProposalComputer.java
```diff
@@ -39,8 +39,12 @@ public FocusedJavaAllProposalComputer() {
 	@Override
 	public List computeCompletionProposals(ContentAssistInvocationContext context, IProgressMonitor monitor) {
 		if (shouldReturnResults()) {
-			List proposals = super.computeCompletionProposals(context, monitor);
-			return FocusedJavaProposalProcessor.getDefault().projectInterestModel(this, proposals);
+			try {
+				List proposals = super.computeCompletionProposals(context, monitor);
+				return FocusedJavaProposalProcessor.getDefault().projectInterestModel(this, proposals);
+			} catch (Exception e) {
+				return Collections.emptyList();
```
ruspl-afed, line 46: I don't know when this exception occurs. Should we just ignore it or log as well?

BeckerFrank: > I don't know when this exception occurs. Should we just ignore it or log as well?

This exception can occur with invalide java code, please see #370.
Example NPE in org.eclipse.jdt.internal.compiler.ast.*

So I think that logging this type of error is of no use.

ruspl-afed: Thank you for confirming @BeckerFrank

ruspl-afed (approved):
LGTM

Issue #535, opened 2024-05-26
Switch to Tycho 4.8.0


ruspl-afed:
[Tycho 4.8.0](https://github.com/eclipse-tycho/tycho/blob/tycho-4.0.8/RELEASE_NOTES.md) contains a number of important changes that we should use for Mylyn releng.

PR #536, opened 2024-05-26
[#535] Switch to Tycho 4.8.0

ruspl-afed:
* switch extensions
* switch parent pom


BeckerFrank:
Because we have a property tycho-version in the two parent pom files the continuous-integration jobs fail.
Should I provide a fix?

ruspl-afed:
> Because we have a property tycho-version in the two parent pom files the continuous-integration jobs fail. Should I provide a fix?

Sorry @BeckerFrank I missed your comment somehow

PR #537, opened 2024-05-26
use ce Version in gitlabImageName

BeckerFrank:
Sorry this was a type of none Mac builds.


gnl42:
I came across that trying to build on Windows wsl :) Thanks

Could you also bump the Jenkins version?   
Several of the plugins complained that 2.375.2 was to old and they needed at least 2.387.1

gnl42 (approved):

BeckerFrank:
> I came across that trying to build on Windows wsl :) Thanks
> 
> Could you also bump the Jenkins version?  Several of the plugins complained that 2.375.2 was to old and they needed at least 2.387.1

Shold we use latest-jdk11 instead?

PR #543, opened 2024-06-01
fix the workspace errors for a new clean instance

BeckerFrank:
we have 6 errors in the workspace for a new instance.
This commit fix the errors using more oomph setup tasks.

### DIFF
--- a/mylyn.docs/wikitext/ui/org.eclipse.mylyn.wikitext.asciidoc.ui/.project
+++ b/mylyn.docs/wikitext/ui/org.eclipse.mylyn.wikitext.asciidoc.ui/.project
@@ -21,11 +21,6 @@
 			</arguments>
 		</buildCommand>
 		<buildCommand>
-			<name>org.eclipse.m2e.core.maven2Builder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
-		<buildCommand>
 			<name>org.eclipse.ui.externaltools.ExternalToolBuilder</name>
 			<triggers>full,incremental,</triggers>
 			<arguments>
@@ -40,6 +35,11 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.pde.PluginNature</nature>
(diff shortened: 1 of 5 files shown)


gnl42 (approved):

merks:
I still have these 4 errors from a fresh clean install:

image

It's strange because navigation find the correct artifacts in the workspace.

I'll investigate...

BeckerFrank:
> I still have these 4 errors from a fresh clean install:
> 
> image
> 
> It's strange because navigation find the correct artifacts in the workspace.
> 
> I'll investigate...

Can you look if the new "Launch mylyn install" was executed during setup.

merks:
It is.  But I have an idea.  Note all my paths are like this:

image

Because very often need to test in a way that simulates a brand new users, I have this in my eclipse.ini:
```
-Duser.home=D:\Users\test35
```

(I need to test in a way that cannot corrupt my actual development space with test results, e.g., locally build unsigned content should not end up in my shared bundle pool.)

So I think the launched maven build task is not producing results in the expected location because the computed user will likely be wrong.   I expect that specifying this

image

will ensure that the user.home of the launch is the same as the user.home of the IDE from which it is launched.

I know almost no one will be affected by this nor needs this, but it also does no harm if you guys don't mind?

merks:
Also, while I'm twisting the world to conform to my needs, do you mind if we turn on this toolbar contribution?

image

BeckerFrank:
Maybe the problem is that you use an other location for the local maven repository. This was the reason for me to add the "Launch mylyn install" .

gnl42:
I would love to have a couple of third party plugins installed by default as well:

- AnyEditTools
- Jeeeyul's PDE-Tools

merks:
That did the trick.  

That being said, it's still disconcerting to me that maven classpath is bound to the jar and not to the appropriate workspace project.

image

merks:
@gnl42 

I see many projects do install what appear to be own **personal** favorite tools...  I'm always kind of disconcerted when I have to accept a license or accept unsigned content just to contribute to some other project.

So let me ask a question: Do you want these installed in every installation, not just your Mylyn development IDE?   Probably so, just as I want (**need**!) viplugin installed everywhere.  As such, using Navigate -> Open Setup -> User I created tasks there; specifically a task to install viplugin and a task to copy the license file to the necessary location where viplugin looks for it:

image

Would that suit your personal needs or are these tools the project itself needs for effective development?

gnl42:
Hi @merks The user.setup sounds perfect for what I want. But I have no idea where it lives or how to set it up.
Can you point me to some documentation please?

gnl42:
But could you change the console settings while you're at it?:

- unlimited size
- tab size 4

merks:
From the menu bar use Navigate -> Open Setup -> User. 

https://wiki.eclipse.org/Eclipse_Oomph_Authoring#Understanding_Setup_Tasks_and_Scopes

Oomph's handy toolbar's right button does this too.

Issue #544, opened 2024-06-01
Mylyn build failures with timeout


ruspl-afed:
We have repeatable timeout failures first appeared from https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/311/
Perhaps it is related with switch to Tycho 4.0.8, but I cannot state it.

These failures puts our final contribution to SimRel 2024-06 at risk

merks:
yes, I think the timeout is too small:

BeckerFrank:
I add another 10 minutes to the timeout option, so now we have 60 minutes.
The strange thing is that I have never seen the timeout in pull requests.

merks:
PR builds don't do signing.  :stuck_out_tongue_winking_eye:

merks:
At 56min 12s it's still close to the 60min threshold where a slow instance can still timeout, and that's after waiting a whole hour already...

BeckerFrank:
new limit set with https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/547

ruspl-afed:
Build statistics looks much better now, thanks @BeckerFrank and @merks

PR #545, opened 2024-06-01
fix: change timeout to 60 Minutes

BeckerFrank:


merks:
It good to link PR to the issue:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/544

PR #546, opened 2024-06-01
Improve the Oomph setup and m2e launchers

merks:
- Specify property user.home=${system_property:user.home} for the m2e launchers so that the user.home of the launched process necessarily matches the user.home of the launching application.
- Add a preference task to show Oomph's tiny two-button setup tool bar.
- Set the eclipse.target.platform variable to None to avoid a prompt for something not intended to be used.
- Set IDs on tasks that are reference to avoid references that are positional which might be corrupted by textual editing.
- Configure clones configure new branches to rebase.
- Configure core/autocrlf to input.
- Set the console to be of unlimited outputsize with tab width of 4.

### DIFF
--- a/org.eclipse.mylyn.releng/launcher/mylyn clean.launch
+++ b/org.eclipse.mylyn.releng/launcher/mylyn clean.launch
@@ -10,6 +10,7 @@
         <listEntry value="maven.test.failure.ignore=true"/>
         <listEntry value="maven.test.error.ignore=true"/>
         <listEntry value="dash.fail=false"/>
+        <listEntry value="user.home=${system_property:user.home}"/>
     </listAttribute>
     <stringAttribute key="M2_RUNTIME" value="EMBEDDED"/>
     <booleanAttribute key="M2_SKIP_TESTS" value="false"/>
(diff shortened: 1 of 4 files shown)


merks:
It strikes me that this setting is just asking for trouble:

image

With this setting it's easy to accidentally commit text files with `\r\n` into the git repository.

Are you folks sure that this would not be better set to `input` rather than `false`?  That way `\n` is not converted to `\r\n` on checkout, but `\r\n` is converted to `\n` on check-in (input)...

merks:
I'm out for a while now....

BeckerFrank:
> It strikes me that this setting is just asking for trouble:
> 
> image
> 
> With this setting it's easy to accidentally commit text files with `\r\n` into the git repository.
> 
> Are you folks sure that this would not be better set to `input` rather than `false`? That way `\n` is not converted to `\r\n` on checkout, but `\r\n` is converted to `\n` on check-in (input)...

Yes we should change this setting! Do you add an other commit to your branch?

merks:
I’ll amend and force push when I get home.

merks:
I believe all these files are committed with wrong line endings:

image

Just download the raw file and have a look:

https://raw.githubusercontent.com/eclipse-mylyn/org.eclipse.mylyn/main/mylyn.reviews/tbr/org.eclipse.mylyn.reviews.tasks.dsl/.classpath

I can create a PR to fix this once this is done...

merks:
This PR includes the requested change from:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/543#issuecomment-2143532560

BeckerFrank (approved):
LGTM

merks:
I think there is nothing controversial here and most certainly nothing that can't be changed.

PR #547, opened 2024-06-01
[#544] set timeout to 80 Minutes

BeckerFrank:
Add more time to the timeout value as we are getting too close to the previous value.


gnl42 (approved):

merks:
Better conservative than sorry!

PR #548, opened 2024-06-01
Correct a files with \r\n in the repository to use purely \n

merks:


merks:
This is hard to review because it shows everything as different even though only the line endings change.

BeckerFrank (approved):
I assume that you use a tool so we can continue and merge

merks:
Yes I can recreate this in minutes. So don’t let it interfere with important work. It’s not high priority!

PR #549, opened 2024-06-02
small oomph setup fixes

BeckerFrank:
1. remove the not working .github repository setup
2. add all Doc projects to the Doc Workingset

### DIFF
--- a/org.eclipse.mylyn.releng/oomph/Mylyn.setup
+++ b/org.eclipse.mylyn.releng/oomph/Mylyn.setup
@@ -402,21 +402,14 @@
       </workingSet>
     </setupTask>
     <setupTask
-        xsi:type="projects:ProjectsBuildTask"
-        id="build.all"
-        refresh="true"
-        clean="true">
-      <description>Build All</description>
-    </setupTask>
+        xsi:type="launching:LaunchTask"
+        id="launch.install"
+        predecessor="mylyn.workingsets"
+        launcher="mylyn install"/>
     <setupTask
         xsi:type="maven:MavenUpdateTask"
         id="maven.update"
         predecessor="launch.install"/>
-    <setupTask
-        xsi:type="launching:LaunchTask"
-        id="launch.install"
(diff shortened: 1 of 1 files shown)


merks:
Instead of removing it, might it be better to fix it?

BeckerFrank:
> Instead of removing it, might it be better to fix it? (I can help fix it.)

I do not really use / need the [.github](https://github.com/eclipse-mylyn/.github) repository. 

I get an java.lang.Exception: org.eclipse.jgit.errors.RepositoryNotFoundException: repository not found: /Users/frank/eclipse/mylyn-admin-test/git/.github

I think we have a git clone problem with repositories starting with a dot.

merks:
Yes, they generally need a bit of special attention...  E.g., like this for simrel:

https://github.com/eclipse-simrel/simrel.build/blob/bcf05da46c2f46508118643655378420688296aa/SimRel.setup#L188-L192

So that it creates a folder simrel.github and not .github.  I can help make it work nicely if you wish...

merks:
BTW, I also think this new build installer tasks does not need to run on each startup and there is a property to disable that...  It's a bit annoying to wait for a build each time you start the IDE...

BeckerFrank:
> Yes, they generally need a bit of special attention... E.g., like this for simrel:
> 
> https://github.com/eclipse-simrel/simrel.build/blob/bcf05da46c2f46508118643655378420688296aa/SimRel.setup#L188-L192
> 
> So that it creates a folder simrel.github and not .github. I can help make it work nicely if you wish...

So the trick is to set the locationQualifier.  Right?

BeckerFrank:
@merks: Please check if this is better now.

merks (approved):
It looks better.

Once it's committed I'll try another completely-from-scratch test run.

I'd suggest committing it without waiting for others to review because there's nothing really controversial, it has has no impact on the build, and I can fully test it better once it's committed.

ruspl-afed (approved):
LGTM

Issue #552, opened 2024-06-06
Eliminate long maven build from the setup


merks:
I believe the maven build was added to address the problem with the errors in the org.eclipse.mylyn.wikitext.maven project.  Which made me wonder why an artifact in the .m2 cache is needed at all.  So I asked about here:

https://github.com/eclipse-m2e/m2e-core/discussions/1777#discussioncomment-9685967

Based on that answer, it seems to me that we can just add the m2e nature/builder to the two projects such that the two org.eclipse.mylyn.wikitext.maven  dependencies resolve to the workspace projects.

I can experiment further to verify that this fixes the problem.  I just wanted to confirm that there is no some other reason for doing a full build (which takes really long which is really quite annoying).

ruspl-afed:
Thanks @merks please proceed with replacing full cli mvn build with what m2e can offer

merks:
I've tested that this works well, but will retest after the commit.  Note I've just disabled the build so that one could locally enable it if desired...

wimjongman:
Can this be closed?

PR #553, opened 2024-06-06
Eliminate long maven build from the setup

merks:
- Ensure that the two projects referenced as dependency by org.eclipse.mylyn.wikitext.maven have an m2e nature and build so that the dependencies resolve to those workspace projects.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/552

### DIFF
--- a/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.textile/.project
+++ b/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.textile/.project
@@ -25,10 +25,16 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.pde.PluginNature</nature>
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.pde.api.tools.apiAnalysisNature</nature>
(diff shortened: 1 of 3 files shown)


merks:
@ruspl-afed 

Maybe there's a reason to avoid doing a commit during the current post-rc2 period?

ruspl-afed:
TL; DR;  please proceed

Speaking from the Change Management perspective we have a release tag https://github.com/eclipse-mylyn/org.eclipse.mylyn/releases/tag/R_4_3_0 and we can create a branch to hotfix 4.3.0 is something will go really wrong.

Speaking from the Release Management perspective we don't have resources to do testing for such a notable source base. We have only a few part time enthusiast that can sporadically fix something. We fully rely on automated tests we have. And the recent degradation of testing infra is a serious risk for us.

I tried to draw attention of Eclipse IDE WG to this problem [here](https://www.eclipse.org/lists/eclipse-ide-wg/msg00354.html) but without too much success. 
Also, during the recent Eclipse IDE - Developers community call,  I asked presenters from the one of Eclipse IDE WG Platinum members if they plan to contribute to Eclipse Mylyn since they use it. The answer was negative.

Well, if vendors are not interested to help Eclipse Mylyn, we can only continue our usual project reanimation protocols as we do already for a several releases. 

Thank you for reading.

merks:
The story for Mylyn is not so different from that for BIRT, where there are a large number of people who want stuff and use stuff, from very, very large companies with lots of money.  But no one is wiling to do much of anything, except maybe complain.  😱 

The whining about wanting BIRT published to Maven central is particularly comical.

The story is pretty much the same for a great many projects...

PR #554, opened 2024-06-08
oomph setup with github fork remote URL

BeckerFrank:
I also tested that #553 works

### DIFF
--- a/org.eclipse.mylyn.releng/oomph/Mylyn.setup
+++ b/org.eclipse.mylyn.releng/oomph/Mylyn.setup
@@ -4,7 +4,6 @@
     xmlns:xmi="http://www.omg.org/XMI"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns:git="http://www.eclipse.org/oomph/setup/git/1.0"
-    xmlns:launching="http://www.eclipse.org/oomph/setup/launching/1.0"
     xmlns:maven="http://www.eclipse.org/oomph/setup/maven/1.0"
     xmlns:pde="http://www.eclipse.org/oomph/setup/pde/1.0"
     xmlns:predicates="http://www.eclipse.org/oomph/predicates/1.0"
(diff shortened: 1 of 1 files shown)


merks:
FYI, I have not documented or advertised the following cool feature...

In order to avoid all this authoring complexity for the Git clone tasks that we see here, Oomph (as of last March) supports a Git Configuration task.
```
<?xml version="1.0" encoding="UTF-8"?>
<git:GitConfigurationTask
    xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:git="http://www.eclipse.org/oomph/setup/git/1.0"
    xsi:schemaLocation="http://www.eclipse.org/oomph/setup/git/1.0 https://raw.githubusercontent.com/eclipse-oomph/oomph/master/setups/models/Git.ecore"
    remoteURIPattern="(?:https://github.com/|git@github.com:)[^/]*(?&lt;!merks)(/.*)">
  <configSections
      name="remote">
    <subsections
        name="merks">
      <properties
          key="fetch"
          value="+refs/heads/*:refs/remotes/merks/*"
          force="true"/>
      <properties
          key="url"
          value="git@github.com:merks$1"/>
    </subsections>
  </configSections>
</git:GitConfigurationTask>
```

You can literally copy the above text and paste it into your Users.setup as I have here as a starting point:

image

The idea is that you can write a pattern that matches the URI of some other arbitrary Git clone task's URI that you wish to "decorate/augment" with your own personal configuration details.  So here I pattern match any GitHub URI, except those from my personal org with the negative look-behind, capture the name of that repository, and map it to a fork with the same name in my personal org.  You can see in the above that the SWT clone will have my remote added to it.

This way I don't have to do it manually, and I don't have to ask the other setup authors to do this for me because it's quite a bit of complex detail to author.  Also, I don't want to have long discussions about the best work flow while attempting to enforce my own idea of the best one...

If you change `merks` to `BeckerFrank` that should do the trick.  And then it should work with any other project's Git clone tasks too...

What do you think?

BeckerFrank:
> FYI, I have not documented or advertised the following cool feature...
> 
> In order to avoid all this authoring complexity for the Git clone tasks that we see here, Oomph (as of last March) supports a Git Configuration task.
>  ...
> What do you think?

I have tested this and it works after adding
```
  <setupTask
      xsi:type="git:GitConfigurationTask"
      remoteURIPattern="(?:https://github.com/|git@github.com:)[^/]*(?&lt;!BeckerFrank)(/.*)">
    <configSections
        name="remote">
      <subsections
          name="BeckerFrank">
        <properties
            key="fetch"
            value="+refs/heads/*:refs/remotes/BeckerFrank/*"
            force="true"/>
        <properties
            key="url"
            value="git@github.com:BeckerFrank$1"/>
      </subsections>
    </configSections>
  </setupTask>
```
and adding `-Duser.home=/Users/frank/eclipse_test` to the Eclipse\ Installer\ Test.app/Contents/Eclipse/eclipse-inst.ini and add an other setup task into the Users.setup
```
  <setupTask
      xsi:type="setup:EclipseIniTask"
      id="user.home"
      option="-Duser.home="
      value="/Users/frank/eclipse_test"
      vm="true">
    <description>Set the User Home Dir</description>
  </setupTask>
```

I now have a compete isolated new Eclipse Test environment. 

So I think we should close this Pull Request.

merks:
I really should do a better job documenting things.  I'm a little torn now because the wiki is on the way out, and I've just not had time to do migration of documentation, what little of it there is.  😱 

Another FYI...

I keep my installer in a permanent location (a nightly installer so that it installs the very latest version of Oomph into everything):

https://wiki.eclipse.org/Oomph#Eclipse_Installer

I modify the eclipse-inst.ini to add these -vmargs:
```
-Duser.home=D:/Users/merks
-Doomph.setup.user.home.redirect=true
```
This is to keep my real user home clean.  The `-Doomph.setup.user.home.redirect=true` directs the installer to generate a '-Duser-home=...' in every installation using exactly the value of the user.home system property of the installer...

BeckerFrank:
@merks: Thanks I now have an new separate clean installation space.

I only see a small problem if you want to fork the repository .github from 2 or more organizations. But that is rather unlikely for me at least.

merks:
Yes that repo is annoyingly named.  But tricky patterns could also deal with that.  I think but would need to test that a more specific pattern specified first would do the trick. Or one can edit the result to use the name actually chosen.  Each person might choose a different name anyway so it’s hard to configure generically.

PR #556, opened 2024-06-13
replace I-Build with 4.33-I-builds in target plattform

BeckerFrank:


merks:
You don’t want to use 4.33-I-builds?

merks (approved):

PR #559, opened 2024-06-16
Prepare releng for 4.4.0

BeckerFrank:


merks (approved):
Mass version changes are hard to review.  😝

BeckerFrank:
> Mass version changes are hard to review. 😝

Yes, but if you also have to change the versions for dependencies, it becomes even more difficult.
Maybe we can change some places to work with ranges instead of a fixed version.
Example: bundle-version="[4.4.0,5.0.0)" instead of bundle-version="4.4.0" in the *.MF files

Should I make this change?

merks:
FYI the hamcrest issue this is a glitch.

Yes, I approved the changes.

Issue #521, opened 2024-05-10
Remove guava in wikitext, part 1


gnl42:
Need to replace guava uses in wikitext with native/commons alternatives

gnl42:
So I've run into a bit of snag. There are a couple of guava classes used by wikitext that don't have direct (or even simple ) replacement code. Specifically CharMatcher/Escapers

I'm not sure what's the best approach at this point:

1. Leave Guava jar in place (wikitext code is the only user)?
2. Take the time and effort to come up with custom alternatives?
3. Extract the appropriate classes from Guava and add them to wikitext (with appropriate mods)?

I have a working Poc of the third option, 25 classes in total. It's an ugly hack, but it works. Potentially would allow for a refactor/rewrite in the future.

Thoughts? 
@BeckerFrank @akurtakov @merks @wimjongman

akurtakov:
I would vote for 1. as adding 25 classes of thirdparty code will exclude us from getting bugfixes that might/will happen upstream. 2. is actually smth that has been happening in the last years although a bit too slow.

merks:
3 sounds nasty 🤮
2 sounds like a lot of work and might end up with a similar but unproven amount of new code. 
1 would leave us with the problem we hoped to eliminate. 

I leave decisions to the folks doing the hard development work.

wimjongman:
3 is an implementation of 2. Why do we want to ditch Guava again?

akurtakov:
> 3 is an implementation of 2.

It could be or not. I've been replacing Guava usages with new Java versions API and keeping Guava for the rest (e.g https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/4fd84e30acddcf4c224a4a63b4487119f343aa14 https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/9ad02738c52b58fbcad91743f3db0f1355c7dc18  https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/2f0cc660aa4f8d50a11e37de7fc151b59bbc651d https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/803292b246c65c115333d12b2730a3b9759bb397 and others), in that case - it's a slow reduction of guava usage without coping code to be maintained by the project.

wimjongman:
> in that case - it's a slow reduction of guava 

Sure, that is reasonable. 
George, Are you ok with leaving Guava in for the other unsupported cases?

[...]

PR #530, opened 2024-05-20
Remove guava in wikitext, Part 1 #521

gnl42:
Replace basic guava constructs with Java 17/Apache commons equivalents

PR #562, opened 2024-06-24
Remove guava in wikitext

stephan-herrmann:
+ address core bundle org.eclipse.mylyn.wikitext

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/521

### DIFF
--- a/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.markdown/src/main/java/org/eclipse/mylyn/wikitext/markdown/internal/GfmIdGenerationStrategy.java
+++ b/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.markdown/src/main/java/org/eclipse/mylyn/wikitext/markdown/internal/GfmIdGenerationStrategy.java
@@ -13,20 +13,49 @@
 
 package org.eclipse.mylyn.wikitext.markdown.internal;
 
+import java.util.Arrays;
 import java.util.Locale;
 
 import org.eclipse.mylyn.wikitext.parser.markup.IdGenerationStrategy;
 
-import com.google.common.base.CharMatcher;
-
 public class GfmIdGenerationStrategy extends IdGenerationStrategy {
 
+	enum ScanState {
+		HEAD, WORD, HYPHEN
+	}
 	@Override
 	public String generateId(String headingText) {
 		String id = headingText.toLowerCase(Locale.getDefault());
 		id = id.replaceAll("[^a-z0-9_-]", "-"); //$NON-NLS-1$//$NON-NLS-2$
-		CharMatcher hyphenMatcher = CharMatcher.is('-');
-		id = hyphenMatcher.trimFrom(hyphenMatcher.collapseFrom(id, '-'));
+		ScanState state = ScanState.HEAD;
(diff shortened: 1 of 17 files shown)


stephan-herrmann:
Various strategies applied:
* use basic API like `Character.isWhitespace()` :)
* use streams
* implement minimal replacements
  * XML-escaping (class XML) is inspired from utils we have in o.e.resources as well as jdt.
  * only in GfmIdGenerationStrategy.generateId() a little algo was implemented from scratch
* for the set of `MarkupLanguage` I felt that filtering duplicates should not be necessary if identical language instances are detected using `equals()` and `hashCode()`.
  * this required to adjust one test to force duplicates into a set: same name but different classes

stephan-herrmann:
My goal is to find out if https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1472 could be implemented with wikitext (markdown) and without guava. For that task I don't expect to need any of the \*.ui bundles, as rendering to HTML should be fully sufficient for JDT. In particular no markdown _editing_ is planned.

merks:
This looks promising.

gnl42:
How will this work live with #521?

merks:
> How will this work live with #521?

If I understand correctly, it provides direct implementations that address https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/521#issuecomment-2109504799 for the non-ui parts.

stephan-herrmann:
I admit I found #530 only after submitting this PR :)

So the current PR is much narrower in scope, but succeeds to fully eliminate guava dependency from the core plugin without adding any other dependency.

I'm open to discussions to find the best of all worlds. I'm driven by the fact that JDT needs markdown rendering by September the latest.

ruspl-afed:
@stephan-herrmann thank you for your PR! Perhaps it was my fault to not mention an existing #530 explicitly during [our discussion ](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2429#issuecomment-2180464270)

I would prefer to have #530 merged first, since it was prepared earlier in the scope of #521 and using the same approach as we already have in other parts of Mylyn codebase, and then discuss your change on top of it to reduce the number of required 3rd party libraries. I hope this is fine with you.

> How will this work live with #521?

@gnl42 could you please resolve conflicts for #530 so we can merge it and proceed with further discussion?

stephan-herrmann:
> I would prefer to have #530 merged first, since it was prepared earlier in the scope of #521 and using the same approach as we already have in other parts of Mylyn codebase, and then discuss your change on top of it to reduce the number of required 3rd party libraries. I hope this is fine with you.

Sure, I certainly didn't mean to jump the queue :)

> > How will this work live with #521?
> 
> @gnl42 could you please resolve conflicts for #530 so we can merge it and proceed with further discussion?

If in doubt, some of the more involved changes, that might require a discussion, could perhaps be postponed for later?

gnl42:
#530 has been merged as #565

gnl42:
@stephan-herrmann  Can you recreate your changes against the current main?

stephan-herrmann:
> @stephan-herrmann Can you recreate your changes against the current main?

See new PR #568

stephan-herrmann:
Hopefully #568 will take this to completion.

PR #565, opened 2024-06-30
remove guava in wikitext, part 1 #521

gnl42:
Recreation of #530.

Replace guava constructs with native/apache commons equivalents.

Still have:
- CharMatcher
- Escaper
- Splitter

with no simple alternatives (that I've found)


ruspl-afed (approved):
LGTM

Thanks a lot @gnl42 for reviving this work!

PR #568, opened 2024-07-04
Remove guava in wikitext

stephan-herrmann:
Also avoid new apache dependencies

Fully covers wikitext core

Tests:
+ modify MLPT.getMarkupLanguagesDuplicateNames():
  - need different language classes now
+ skip CommonMarkSpecTest line 6700 for now
+ revert expectation in LineSequenceTest.assertLookAheadFailsFast()

fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/521

Replaces #562 to build on #565


stephan-herrmann:
Some notes on top of https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/562#issuecomment-2187555735 :
* Simple substitutes:
  * `org.eclipse.mylyn.wikitext.util.Preconditions`
  * `org.eclipse.mylyn.wikitext.util.Strings` (caveat: null != empty != blank)
* More involved substitutes:
  * `org.eclipse.mylyn.wikitext.internal.util.XmlUtil`, see https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/562#issuecomment-2187555735
  * `org.eclipse.mylyn.wikitext.util.IDStrategies` generalized from my previous take at `GfmIdGenerationStrategy.generateId()`
    * caveat: clients use different locales for `toLowerCase()`, intentionally?
  * `org.eclipse.mylyn.wikitext.util.UrlUtil` - **here be dragons, see below**
* For `toString()` methods I let JDT propose a default implementation and manually fine tuned it to produce the output expected from tests
* `String.repeat()` does not require any library :)
* `PotentialEmphasisSpan.isWhitespace()`: caveat: whitespace != whitespace (Java vs. unicode)
* `MarkdownDocumentBuilder.emitContent()` caveat: expects ' ' but no other whitespace, intentionally?

For Url escaping, I had hoped that `java.net.URLEncoder` would satisfy all requirements and needs, but alas, it escapes more than what we expect in `CommonMarkSpecTest`.
* `org.eclipse.mylyn.wikitext.util.UrlUtil.escapeUrlFormParameters(String)` seems to be suitable
* `org.eclipse.mylyn.wikitext.util.UrlUtil.escapeUrlFragment(String)` heuristically tries to mimic the expected behaviour
  * `ALLOWED_IN_FRAGMENT` captures legal chars, so we scan to find the first char needing an escape, but from that char onwards, we might escape more than desired
  * see the new line exclusion (6700) in `CommonMarkSpecTest`: 
    * given `[foo<http://example.com/?search=](uri)>`
    * expected `href="http://example.com/?search=%5D(uri)"` -- `(` and `)` expected unescaped but current impl will escape them :warning:

gnl42:
@stephan-herrmann  The ticket is supposed to remove guava references in wikitext, but it removes a lot of non-guava code as well which to me is outside the scope of the ticket. 

It should only concentrate on guava, not muddy the waters with removing apache calls which does not need to be removed.

gnl42 (changes requested):
Should only remove guava uses. Should not remove uses of apache just because it is apache

merks:
> Should only remove guava uses. Should not remove uses of apache just because it is apache

I think "we" should reconsider that.  As I understand it, the more 3rd party dependency one drags in, the more problematic it is to consume because more baggage is dragged in.  One might argue the water is currently very muddy currently.  And it's probably safe to assume @stephan-herrmann didn't do this for new reason at all...

gnl42:
The apache library is already in the build so I don't consider that adding a library, as opposed to adding something to the target file.

Either way, these kind of changes should be in their own issue, not one for guava

merks:
This will be the third time @stephan-herrmann is being asked to rework a contribution. I don't speak for him, but in general eventually folks tend to give up when hurdles appears in the way.  I know Stephan is very persistent though...

One could just as well rename the ticket to change in scope:

> Improve the consumablity and reusability of wikitext by reducing 3rd party dependencies.

I suppose it's none of my business.  Somehow it feels poorly to me....

gnl42:
Removing guava is an very different beast than replacing third party library calls with "inline"/copied equivalents (and when I suggested importing the guava classes into wikitext for the same reason I was shot down along the lines of it creating a maintenance issue tracking the original version).

And as @stephan-herrmann points out in the description, there are already issues with the changes and I see several other issues already.

I could go along with a new ticket for replacing the apache uses (missing Strings.repeat() was a 'my bad' on my part) but I have bad feelings about making this issue a catch all one.

merks:
Yes, I understand your concerns about scope creating and have even more things to review and more risk subtle behavior changes.

stephan-herrmann:
Thanks, @gnl42  and @merks  for commenting.

First things first: we have a definite need for improvement around `UrlUtil.escapeUrlFragment()`. The current implementation does not pass one particular test of  `CommonMarkSpecTest` (see above for disablement). 

* I'm puzzled by the fact that in HEAD `UrlEscapers.urlFragmentEscaper().escape(link)` is used to encode *entire URIs*, not just a URL fragment. Can anyone explain this?
* I wonder if we should use `URL` to first parse the string into parts, which can then be encoded with their specific rules?
* Otherwise we may have to bite the bullet and implement full parsing, i.e., iterate all chars, recognize different URL parts as we go and perform substitution as suites the recognized part.
  * Do you have a definite specification at hand that could drive this implementation? The best I found so far is at the level of blog posts, rather than specification.

Since I'm not an expert in web-programming, I'd highly welcome some input here.

stephan-herrmann:
As for replacing usage of apache utils:
* As said, my goal is to make wikitext consumable by JDT. At the end of the day all transitive dependencies must be covered by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target which lists commons-lang:commons-lang:2.6 but not org.apache.commons.lang3.
  * Can anyone comment on feasibility to add commons.lang3 to eclipse-sdk-prereqs?
* I started from my previous PR where I made direct substitution from guava to own code resp. JDK libraries.
  * Seeing migration from guava to commons.lang3 in [#565](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/565) I figured it best to do the next step right now while it's still ease to compare with the previous state using guava. E.g., migrating to `Validate.isTrue()` changed some `IllegalArgumentException` to `IllegalStateException`, which I thought best to revert immediately, rather than later.
  * Looking specifically at dependencies of `org.eclipse.mylyn.wikitext` I only reverted some additions from  [#565](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/565)
  * Please excuse if I got carried away removing the dependency also in other plugins

Asking pragmatically, which road poses least obstacles?
* add commons.lang3 to eclipse-sdk-prereqs and keep the dependency in wikitext (except for `StringUtils.repeat()` I guess :) ).
* or: split this PR into two PRs to be reviewed separately?
  * more work for me, perhaps even more work for reviewers?
* or: review this PR as is
  * some changes can be best compared against the state before  [#565](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/565)

Is there any specific problem with any of my replacements of apache utilities, or is the concern only about the volume of the change? I don't recall any of the apache replacements as particularly tricky. Changes to `toString()` methods have a high volume, but very low complexity. @gnl42  which parts are you worried about?

And please don't let the discussion about apache distract us from the necessary work around URL escaping!

ruspl-afed:
Thank you for your efforts and for your optimistic attitude @stephan-herrmann ! 
Let me add my two cents to this discussion.

1) The scope of the task to remove guava is big enough and we shouldn't extend it to other topics if we want to converge within a reasonable time.  So, if we want to discuss apache replacement or troubles with particular guava type replacement,  let's postpone questionable items for other PRs so that we can merge the rest of the work. I believe that this is how we can make faster progress.

2)  Nobody says that we _must_ literally reuse existing Mylyn Docs API/implementation in its current state. We may introduce other bundle(s) to be clean from 3rd party libraries and have sufficient interface/implementation to be suitable for _JDT , LSP4E_ and other potential consumers. In this case tasks to remove guava, apache, etc are no longer coupled with a task to have reusable markdown support. So, instead of discussing how to replace the particular 3rd party type we can focus on defining a good interface for markdown parsing suitable for JDT needs. If you think this would be helpful you are very welcome to create a dedicated issue to discuss expected markdown interface capabilities @stephan-herrmann

stephan-herrmann:
> add commons.lang3 to eclipse-sdk-prereqs

@merks with your experience from simrel do you want to comment on the feasibility of this?

stephan-herrmann:
> Nobody says that we must literally reuse existing Mylyn Docs API/implementation in its current state. We may introduce other bundle(s) to be clean from 3rd party libraries and have sufficient interface/implementation to be suitable for JDT , LSP4E and other potential consumers. In this case tasks to remove guava, apache, etc are no longer coupled with a task to have reusable markdown support. So, instead of discussing how to replace the particular 3rd party type we can focus on defining a good interface for markdown parsing suitable for JDT needs. If you think this would be helpful you are very welcome to create a dedicated issue to discuss expected markdown interface capabilities @stephan-herrmann

Thanks @ruspl-afed  for considering the big picture. I haven't yet started to work on the actual interface from JDT, but from other use of wikitext I'm confident that just the core plugins should be useful without new API. That's why I'm focusing on these three bundles:
* org.eclipse.mylyn.wikitext
* org.eclipse.mylyn.wikitext.commonmark
* org.eclipse.mylyn.wikitext.markdown -- _just to have the option to select the best matching flavor_

If you look at this PR, those 3 plugins are all I intended to modify (plus corresponding tests).

The only change outside that scope is in [ScreenshotOnTimeoutRule.java](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/568/files#diff-fccd6cda7b7796dfc68a38f582de81952e6cb91ce7fe9cd2c6b6d069ce5b8c5e) where I was driven by my (overzealous?) removal of WikiToStringStyle. 
* this part could be reverted fairly easily
* OTOH, that code change (in `toString()`!) really isn't rocket science and shouldn't require more then 10 sec to review.

merks:
I foresee no problem at all adding this to the Platform's *.target
```
        <dependency>
          <groupId>org.apache.commons</groupId>
          <artifactId>commons-lang3</artifactId>
          <version>3.14.0</version>
          <type>jar</type>
        </dependency>
```
The following appears to be there for tests only:
```
			  <dependency>
				  <groupId>commons-lang</groupId>
				  <artifactId>commons-lang</artifactId>
				  <version>2.6</version>
				  <type>jar</type>
			  </dependency>
```

It's not actually in the p2 repository:

image

stephan-herrmann:
> I foresee no problem at all adding this to the Platform's *.target

Thanks @merks, I really was afraid that would cause more trouble, but your statement is encouraging.

Please find in [bb20f51](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/568/commits/bb20f51c5f39a4e2b690190dba7fb21d4e378ed8) a proposed compromise where I reduced the surface area by reinstating some use of commons-lang3.

Note that some of my remaining changes were done before I even saw  #565, where certain use of commons.lang3 was introduced. Specifically I believe it is useful to maintain the distinction between checkState vs. checkArgument (throwing different exceptions - as it was before #565), rather than mapping both to the same `Validate.isTrue()`. (btw, why not `org.eclipse.core.runtime.Assert.isTrue()`?). Also replacing guava's trivial `isNullOrEmpty()` with a trivial same-named implementation of our own didn't look like causing pain for any reviewer. We need our own `Strings` class anyway for replacements of `CharMatcher` functionality like `isBlank()`.

@gnl42  if you still see issues, please be specific, or: feel free to add more reverts to the PR.

At this point I can only repeat: 
> And please don't let the discussion about apache distract us from the necessary work around URL escaping!

I would have loved to spend time on that rather on reverting changes.

gnl42:
@stephan-herrmann I still think it should be two issues. but ...
I don't understand the desire to remove the third party libraries but it sounds like it's something to do with the wikitext setup in which case I apologize. Still not happy about adding our own version of the library versions.

gnl42 (approved):

stephan-herrmann:
Thanks @gnl42 for approval.

Incidentally just now we found a leaner solution in orbit: [org.commonmark_0.22.0.v20240316-0700.jar](https://www.eclipse.org/downloads/download.php?file=/tools/orbit/simrel/orbit-aggregation/milestone/S202406281210/plugins/org.commonmark_0.22.0.v20240316-0700.jar).

Unless we find some deficiencies in that library (which is unlikely, the javadoc reference implementation uses the same), JDT no longer needs mylyn wikitext for Java 23 support.

I hope this PR didn't stir up too much unnecessary dust, but perhaps it still helps other potential consumers like lsp4e. Feel free to modify the PR as needed.

mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext/src/main/java/org/eclipse/mylyn/wikitext/util/IDStrategies.java
```diff
@@ -0,0 +1,70 @@
+/*******************************************************************************
```
gnl42, line 1: For what it's worth, Github Copilot came up with this solution:

```
			return id.toLowerCase(Locale.ENGLISH)
				.replaceAll("[^a-z0-9_]", "-")
				.replaceAll("-+", "-")
				.replaceAll("^-|-$", "");
```

stephan-herrmann: Looks nice :)

Could performance be an issue with all those regular expressions (patterns compiled anew for each invocation)?

gnl42:
Maybe, if so:
```
	private static final Pattern NOT_XXX_NNN = Pattern.compile("[^a-z0-9_]"); //$NON-NLS-1$

	private static final Pattern HYPHEN = Pattern.compile("-+"); //$NON-NLS-1$

	private static final Pattern HYPHEN_START_END = Pattern.compile("^-|-$"); //$NON-NLS-1$

	@Override
	public String generateId(String id) {
		id = id.toLowerCase(Locale.ENGLISH);
		id = NOT_XXX_NNN.matcher(id).replaceAll("-"); //$NON-NLS-1$
		id = HYPHEN.matcher(id).replaceAll("-"); //$NON-NLS-1$
		id = HYPHEN_START_END.matcher(id).replaceAll(""); //$NON-NLS-1$

		return id;

	}
```

merks:
@gnl42 

I've recently been doing performance tuning on a client application and have done changes exactly like this.   It's less elegant and pretty, but definitely performs better!

gnl42:
> I've recently been doing performance tuning on a client application and have done changes exactly like this. It's less elegant and pretty, but definitely performs better!

I compared the runtime of the various versions (10000000 calls each version):
```
Id="--e4bc9bef-0238-4f87-813d-f0b78bdb5c21--"

Guava took 443ns average per call
State machine took 286ns average per call
Regex took 1041ns average per call
Compiled regex took 788ns average per call
```

In the end it basically boils down to how often is the method called vs. clarity of code

gnl42:
There are still some references to guava in the pom files and friends.

Easier to create a mylyn branch to clean up

PR #569, opened 2024-07-06
Use org.glassfish.hk2.osgi-resource-locator from latest Orbit

merks:
- Use version 1.0.3 which is newer than all the current 2.x versions

https://github.com/eclipse-simrel/simrel.build/issues/438

The 2.5.0 versions are all older than the 1.0.3 version:

https://repo1.maven.org/maven2/org/glassfish/hk2/osgi-resource-locator/


merks:
@gnl42  @BeckerFrank @ruspl-afed 

Guys, I assume you have no concerns about this trivial change...

ruspl-afed (approved):
LGTM
However having 1.0.3 newer than any 2.x is not so trivial

merks:
> LGTM However having 1.0.3 newer than any 2.x is not so trivial

I didn't actually test what the build produces, only how the *.target resolves in the IDE.  How Tycho does it versus how PDE does it might be two different answers.  I shall check now locally...

merks:
I tested at locally it has the 1.0.3 version in the repository:

ruspl-afed:
> Is there some other concern I've overlooked?

None from my side. Thank you @merks

merks:
Very good.  I will given the other folks a bit more time, but I don't think there's anything controversial here.  😀

PR #572, opened 2024-07-11
Ensure EGit/JGit 7.x is in the permitted version range

merks:


merks:
@BeckerFrank  @ruspl-afed 

EGit/JGit will move to a 7.x version so we need to be sure that works with Mylyn:

https://github.com/eclipse-simrel/simrel.build/pull/458

ruspl-afed:
Thank you for creating this PR @merks , let's wait if Mylyn can compile with EGit 7

merks:
Yes, I've marked it as a draft that that is doesn't get merged, i.e., I don't want to make this specific target platform change permanent.

merks:
Note the build worked:

image

I'll revert the target platform change.

merks:
@ruspl-afed 

I think we could merge this.  OK?

If there is a new build, I can test that EGit and Mylyn work together in SimRel...

ruspl-afed (approved):
LGTM

PR #573, opened 2024-07-11
Fix dash project github org

akurtakov:
To fix path to licensecheck gh action.


akurtakov:
No need to wait for jenkins as it's change in gh action setup only.

Issue #578, opened 2024-07-15
org.eclipse.pde.internal.core.natures.PDE missing


gnl42:
Suddenly seeing complaint about org.eclipse.pde.internal.core.natures.PDE.PLUGIN_NATURE missing from org.eclipse.pde.core_3.18.200.v20240714-0803

![image](https://github.com/user-attachments/assets/ee10ccae-23c3-493c-8097-532ea798765d)

merks:
Where do you see this complaint?  Here are how PDE declares the natures:

![image](https://github.com/user-attachments/assets/fb707f75-5edc-4de4-ae70-f872bda20143)

So all the nature classes appear to be present.  

I'm not sure to what `org.eclipse.pde.internal.core.natures.PDE` and `org.eclipse.pde.internal.core.natures.PDE.PLUGIN_NATURE` refer.

It seems some constants have been moved here recently:

![image](https://github.com/user-attachments/assets/c208b111-d9a1-4ce4-8075-0454a41665a0)

So the have been moved:

![image](https://github.com/user-attachments/assets/dc171f31-1312-4b52-b436-96cc54e76972)

This sounds like a strange variation of your SWT problem where we've not yet established what exactly you have installed:

https://github.com/eclipse-platform/eclipse.platform.swt/issues/1343#issuecomment-2227207669   

Somehow you don't have an installation that's consistently 2024-09 M1, I guess.  But how and why?  I've update more than a dozen IDEs today without a problem. 

Is this problem in a new installation or an updated installation?

gnl42:
This is a brand new installation (last night)

merks:
Can you share the installation details to show which version of pde is installed?   Also details about where the complaint appears might help.  The more details about what you see and how you got to that state would help.

gnl42:
Hmm, 
  I'm seeing the same problem with a fresh install on my Mac.

Windows 11:
![image](https://github.com/user-attachments/assets/9ce1c461-918b-47b8-894b-f2eb5390d77e)

MacOs:
![image](https://github.com/user-attachments/assets/5734d8b7-10aa-4e64-a541-97c5c6d9bfe5)


```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.8:compile (default-compile) on project org.eclipse.mylyn.context.sdk.java: Compilation failure: Compilation failure: 
[...]

PR #582, opened 2024-07-16
Use IBundleProjectDescription.PLUGIN_NATURE

merks:
- Avoid PluginProject.NATURE because that was very recently added there.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/578


merks:
@ruspl-afed 

FYI, this is a better fix.

gnl42 (approved):
Looks better to me

ruspl-afed (approved):
LGTM
Thank you so much @merks !

PR #584, opened 2024-07-17
577 replace orgapachecommonslang3stringescapeutils with orgapachecommonstextstringescapeutils

gnl42:
lang3.StringEscapeUtils was deprecated in 3.6 and replaced by text.StringEscapeUtils

Eclipse is showing an error complaining about 'The method ... has been removed' and I have no idea where that is coming from or how to get rid of if?

image

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
+++ b/mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
@@ -14,7 +14,8 @@ Export-Package: org.eclipse.mylyn.commons.core,
  org.eclipse.mylyn.commons.core.storage;x-internal:=true,
  org.eclipse.mylyn.internal.commons.core;x-internal:=true,
  org.eclipse.mylyn.internal.commons.core.operations;x-internal:=true
-Import-Package: org.apache.commons.collections4.multimap;version="4.4.0"
+Import-Package: org.apache.commons.collections4.multimap;version="4.4.0",
+ org.apache.commons.text;version="[1.12.0,2.0.0)"
 Require-Bundle: org.eclipse.core.net;bundle-version="0.0.0",
  org.eclipse.core.runtime;bundle-version="0.0.0"
 Bundle-ClassPath: .
--- a/mylyn.commons/org.eclipse.mylyn.commons.core/src/org/eclipse/mylyn/commons/core/HtmlStreamTokenizer.java
+++ b/mylyn.commons/org.eclipse.mylyn.commons.core/src/org/eclipse/mylyn/commons/core/HtmlStreamTokenizer.java
@@ -315,6 +315,26 @@ public class HtmlStreamTokenizer {
 	}
 
 	/**
+	 * Returns a string with HTML escapes changed into their corresponding characters.
+	 *
+	 * @deprecated use {@link StringEscapeUtils#unescapeHtml4(String)} instead
+	 */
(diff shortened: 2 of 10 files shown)


akurtakov (approved):
I haven't done detailed testing that behavior is 1:1 the same but overall it looks good to me.

ruspl-afed (approved):
LGTM

merks (approved):
Looks fine.

gnl42:
Could someone check out the branch and see if you get the Eclipse errors I get?

Or is just me (again)

merks:
After resolving the target platform as specified in the PR, I have these API errors:

image

I would suggest that instead of deleting the methods and breaking the API that you mark them deprecated for removal and delegate to StringEscapeUtils to reduce the baggage:
```
	/**
	 * Returns a string with HTML escapes changed into their corresponding characters.
	 * 
	 * @deprecated use {@link StringEscapeUtils#unescapeHtml4(String)} instead
	 */
	@Deprecated(forRemoval = true, since = "4.4.0")
	public static String unescape(String s) {
		return StringEscapeUtils.unescapeHtml4(s);
	}

	/**
	 * Replaces (in-place) HTML escapes in a StringBuffer with their corresponding characters.
	 * 
	 * @deprecated use {@link StringEscapeUtils#unescapeHtml4(String)} instead
	 */
	@Deprecated(forRemoval = true, since = "4.4.0")
	public static StringBuffer unescape(StringBuffer sb) {
		return sb.replace(0, sb.length(), StringEscapeUtils.unescapeHtml4(sb.toString()));
	}
```

merks:
I see this happens in two places:

image

Such a great utility it's good to have two copies.  😱 

This class should have been marked for removal, but probably that's before the days that was supported.

merks:
Of course the alternative is delete these things---in the latter case the entire class---and suppress the API errors.

gnl42:
> Of course the alternative is delete these things---in the latter case the entire class---and suppress the API errors.

I don't understand where the API errors are coming from though? I can't find any references to the methods in the file system so how does Eclipse remember them?

merks:
The API baseline is populated with the contents of the previous release:

image

The API tools compare against that content and notice when you break API by removing a method from a class that is considered API.

It's populated by this in the setup:

image

gnl42:
Cute.  Is there a Dummies book for all of this?  This is all new to me

What makes a method part of an API then? 
And how would the method be removed in the end  without Eclipse complaining?

merks:
There is a quick fix on the errors:

image

It's good to include a comment to describe the reason for the filter:

image

It creates this file which needs to be committed:

image

With Filters... you can show the .* resources.

image

---

The Mylyn history is really messy. This PR is for just one change but it involves 5 commits where the intermediate states are effectively bogus/wrong:

image

Other projects insist on folks squashing commits and rebasing them, amending the commits and force pushing them so that a PR is typically a single commit.  That way one ends up with a clean simple history:

image

ruspl-afed:
> Other projects insist on folks squashing commits and rebasing them, amending the commits and force pushing them so that a PR is typically a single commit. That way one ends up with a clean simple history

Thank you for your feedback @merks , let's try to enforce this for Mylyn as well

merks:
The description here is worth reading if you're not sure how to amend and force push:

https://github.com/orgs/eclipse-simrel/discussions/3

After I wrote that someone mentioned that at the end, it's best to use Rebase and Merge rather than Squash and Merge as I showed in the description above:

image

PR #587, opened 2024-07-19
Update the API baseline setup to use the proper latest feature IDs

merks:
- Use https://download.eclipse.org/mylyn/updates/release/latest and https://download.eclipse.org/releases/latest so that the baseline update to the new release as it becomes available.


merks:
The feature IDs were changes and the updates sites for the API baseline included much older versions.  This uses all the features of the more recent release and update sites that will always provide the latest release.

merks:
Given this has no build impact, I will just merge it.

Issue #611, opened 2024-09-02
Prepare releng to 4.5.0


ruspl-afed:
We need to prepare releng for the next release cycle

PR #613, opened 2024-09-12
change repository url to 4.33 final (#611)

BeckerFrank:


wimjongman:
This repo does not look like it is a final name. @merks WDYT?

ruspl-afed (approved):
For me it does. It has "R-" prefix. May be you are referring to a composite that sounds like "2024-09" @wimjongman  but we prefer to use precise p2 URLs for better dependency control.

merks:
> This repo does not look like it is a final name. @merks WDYT?

In the Oomph setups we even associate a variable with each Platform SDK repository so that one can more easily reuse the URLs:

image

So definitely the one used in this PR is final.

wimjongman:
Thanks for verifying,

Issue #617, opened 2024-09-25
multipass gitlab rule update (add TARGETARCH)


BeckerFrank:
With https://gitlab.com/gitlab-org/omnibus-gitlab/-/issues/8470 we need to modify the rule to add the new arg TARGETARCH

PR #618, opened 2024-09-25
add TARGETARCH to docker image build (#617)

BeckerFrank:


org.eclipse.mylyn.releng/multipass/mylyn_setup.yml
```diff
@@ -125,6 +125,7 @@
 #          - {  servmame: "jenkins-2_303_3", port1: 8081, port2: 50001, data_dir: "jenkins-2.303.3", image_tag: "jenkins/jenkins:2.303.3-lts-jdk11", extra_info: '' }
 #          - {  servmame: "jenkins-1_651_1", port1: 8084, port2: 50004, data_dir: "jenkins-2.303.3", image_tag: "jenkins:1.651.1", extra_info: '"' }
           - {  servmame: "jenkins-latest-jdk11", port1: 8083, port2: 50003, data_dir: "jenkins-2.303.3", image_tag: "jenkins/jenkins:latest-jdk11", extra_info: '"default": "true"' }
+          - {  servmame: "jenkins-latest", port1: 8084, port2: 50004, data_dir: "jenkins-2.303.3", image_tag: "jenkins/jenkins:latest", extra_info: '' }
```
gnl42, line 128: Why two Jenkins instances of the same version?

BeckerFrank: > Why two Jenkins instances of the same version?

No, they are different versions latest-jdk11 is 2.462 and latest is 2.477

But you are right, we only need one. This was just a test on my side and I forgot to remove this. I created an other commit and comment that line out.

gnl42: Sorry, was just looking at the data set version (2.303.3)

gnl42 (approved):
Loos good

Issue #626, opened 2024-10-11
Restore org.eclipse.mylyn.wikitext.context.ui/org.eclipse.mylyn.wikitext.tasks.ui modules to wikitext


gnl42:
Anyone know why these to modules were removed from the wikitext build with 46f52386c5764ebfd51f6859f8101ed436617806? The comment by @leods  seems to imply that it was a guava problem, but since most of guava has been removed now is this still an issue?

I've been trying to add confluence markup support to the JiraConnector with no luck and finally tracked the problem down to the tasks.ui module having been removed.

PR #627, opened 2024-10-12
Restore wikitext.context.ui/wikitext.tasks.ui modules to wikitext #626

gnl42:
Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/issues/626

### DIFF
--- a/mylyn.docs/wikitext/ui/pom.xml
+++ b/mylyn.docs/wikitext/ui/pom.xml
@@ -2,13 +2,13 @@
 <!--
  *******************************************************************************
  * Copyright (c) 2024 Contributors to the Eclipse Foundation
- * 
+ *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * https://www.eclipse.org/legal/epl-2.0/.
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  * Contributors:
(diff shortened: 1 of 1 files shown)


merks:
FYI, this is a dash problem:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/16459#note_2835287

ruspl-afed (approved):
fine with me
however I would wait for dash fix before submitting

ruspl-afed:
Are you sure this will be enough for you to add these modules to the reactor @gnl42 ? 
I assume you expect these bundles to be available with p2 repository, so you may need add them to a feature.

merks:
Note that the dash problem is 10 days old.  At this point it's not clear how long you will need to wait.   In any case, I expect every call to dash will fail this way regardless of what you do or don't do at this point...

gnl42:
> Are you sure this will be enough for you to add these modules to the reactor @gnl42 ? I assume you expect these bundles to be available with p2 repository, so you may need add them to a feature.

Sigh, I'll never get this right. You're right, I missed the feature change in the commit.

I also found a problem with the wikitext ant projects where they were not updated by the setVersion=4.5.0 so they still reference version 4.2

gnl42:
What is dash?

merks:
The thing that runs and opens/reports the iplab issues:

image

PR #647, opened 2024-12-07
Include org.apache.lucene.facet in the update site

merks:
- The 9.12 version of org.apache.lucene.sandbox has a new dependency on that bundle.
- Update the target platform to use the 4.34 release repository.

### DIFF
--- a/org.eclipse.mylyn-site/category.xml
+++ b/org.eclipse.mylyn-site/category.xml
@@ -201,6 +201,7 @@
    <bundle id="org.apache.httpcomponents.httpcore"/>
    <bundle id="org.apache.lucene.analysis-common"/>
    <bundle id="org.apache.lucene.core"/>
+   <bundle id="org.apache.lucene.facet"/>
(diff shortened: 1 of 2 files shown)


merks:
@ruspl-afed 

I was confused when trying to contribute a datatools update why facet is missing even though sandbox requires it.  But now I notice the older version of sandbox in in SimRel:

image

We should fix this for 2025-03 to ensure that the latest versions if all required lucene bundles are aggregated.

ruspl-afed (approved):
Thank you for the fix Ed @merks.
Will you need a Mylyn M0 contribution for SimRel 2025-03? 
If so, when?

merks:
I see there still need to be version increments before producing a new milestone.   There's no big rush.   When Mylyn contribute M1 I can update datatools at that point.

(The datatools release update site has the same problem with missing lucene.facet, but it doesn't really make so much sense to spin an entirely new release only to fix that problem because generally datatools will work with newer versions of lucene contributed by other projects that are building more frequently.)

PR #648, opened 2024-12-07
Use ubuntu instead of centos

merks:
- Fred Gurr recommended changing to ubuntu for other projects of mine.
- Enhance the build description to show details on the ci instance.


ruspl-afed (approved):
Thank you Ed @merks for providing this improvement

PR #651, opened 2024-12-16
Changes to support lucene 10.0.0

merks:
- Update the target platform to use 4.35-I-builds and 2025-03 orbit.
- Because there is API breakage, require minimally lucene 10.

### DIFF
--- a/mylyn.tasks/org.eclipse.mylyn.tasks.core/src/org/eclipse/mylyn/internal/tasks/core/data/TaskDataExternalizer.java
+++ b/mylyn.tasks/org.eclipse.mylyn.tasks.core/src/org/eclipse/mylyn/internal/tasks/core/data/TaskDataExternalizer.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.mylyn.internal.tasks.core.data;
 
+import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
 import java.io.FilterInputStream;
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
```diff
@@ -96,4 +96,5 @@
 			<unit id="org.objenesis" version="0.0.0"/>
 		</location>
 	</locations>
+	<targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
```
merks, line 99: I thought this line would fix the license checking because it helped with the Platform.  But apparently not.

ruspl-afed: May be we need to pass a parameter to fulfill this https://github.com/eclipse-dash/dash-licenses/blob/bc40f27ca7543063a6fc99d97bc308dd0b568f9c/.github/workflows/mavenLicenseCheck.yml#L46

merks: Yes, that would make sense.

I'm not sure I understand the test failures either.  Maybe it's more of a Java 21 thing.

I don't have further time to help on this. I hope to leave it your guys' capable hands.  Okay?

ruspl-afed: Unfortunately I have no idea how to add more commits to the PR, that was created from the fork of another person. 
Is this possible at all or should I open another PR instead?

merks: Don’t worry. Just close this and create a better pr. I won’t be the slightest bit offended!!

gnl42: Looks like you can do it with this: https://stackoverflow.com/questions/44030176/how-to-modify-someone-elses-github-pull-request

merks:
I did some debugging and it seems that Java 17's transformer happily produces invalid XML 1.0:

image

Even encoded as entities, those are not valid characters:

image

Serialization would have to use XML 1.1 for that to be valid:

image

So the tests that are failing are testing based on the apparently bogus assumption that one can serialize characters in the range of \u0001 - \u001F valid which in Java 21 is no longer the case.

mylyn.tasks/org.eclipse.mylyn.tasks.core/src/org/eclipse/mylyn/internal/tasks/core/data/TaskDataExternalizer.java
```diff
@@ -193,19 +194,34 @@ public TaskDataState readState(InputStream in) throws IOException, SAXException
...
 			writer.write(state);
+			out.write(temp.toByteArray());
 		} catch (TransformerException e) {
 			throw new IOException("Error writing task data" + e.getMessageAndLocation()); //$NON-NLS-1$
 		} catch (SAXException e) {
-			throw new IOException("Error writing task data" + e.getMessage()); //$NON-NLS-1$
+			// This condition is just like the one in org.eclipse.mylyn.internal.tasks.core.data.TaskDataStore.readState(File)
+			if (!xml11 && e.getMessage() != null && (e.getMessage().contains("invalid XML character") //$NON-NLS-1$
+					|| e.getMessage().contains(" \"&#"))) { //$NON-NLS-1$
+				writeState(out, state, true);
```
merks, line 221: FYI, this rather ugly hack is analogous to the rather ugly hack when reading the content:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/e56a84cab82f565caf21cb3d3ea1f9ef33ca8795/mylyn.tasks/org.eclipse.mylyn.tasks.core/src/org/eclipse/mylyn/internal/tasks/core/data/TaskDataStore.java#L134-L145

Given all the testing around these invalid characters it's probably best to assume they are important to handle correctly.

merks:
@ruspl-afed @gnl42 

Are you okay with these changes.  It's a bit ironic to have these problems with Java 21 when Mylyn has been primarily used in IDEs running with Java 21 for quite some time...

ruspl-afed (approved):
Thank you for investing your time @merks 
LGTM
If people are not fully happy with your solution, they can always propose improvement.

PR #652, opened 2024-12-17
rework single and container and prepare for setup a new server reached from the internet

BeckerFrank:
Here is what I did.

- create playbooks to save, reload and update the pulled docker Images so that you do not need to pull them every time you recreate the environment.
- adjust the Bugzilla setup to match the latest repository branch names
- rework and then the setup for mode single and container.
  Swarm mode is not really tested an can have some issues.
- Multipass Instance can now have an local DNS Address so that you can share the instance with the internet.
  you have to change some settings in the inventory


wimjongman (approved):
As far as I can understand, this looks good to me. Thanks, Frank!

ruspl-afed (approved):
Thank you for these improvements @BeckerFrank

PR #655, opened 2024-12-23
Update to Tycho 4.0.10

merks:
- Update the justj.p2.manager to use Java 21 because eventually it will need Java 21.


ruspl-afed (approved):
Not sure how Tycho 4.0.10 related with Java 21 for JustJ, but both changes LGTM
Thank you @merks

merks:
It was a two-for-one deal. 😜

Issue #656, opened 2024-12-24
Switch to JavaSE-21


gnl42:
With #655, Java 21 is now the default  so update the various references to 17

PR #659, opened 2024-12-29
switch to javase 21 #656

gnl42:
Change the compiler to JavaSE-21 from JavaSE-17


mylyn.commons/org.eclipse.mylyn.commons.ui.tests/src/org/eclipse/mylyn/commons/ui/ShellDragSupportTest.java
```diff
@@ -40,6 +41,7 @@ public class ShellDragSupportTest {
 	public void setUp() {
 		shell.setLocation(100, 150);
 		when(composite.getShell()).thenReturn(shell);
+		when(composite.getDisplay()).thenReturn(Display.getDefault());
```
ruspl-afed, line 44: looks good

org.eclipse.mylyn.releng/bin/synch_code_style.sh
```diff
@@ -34,6 +34,7 @@ for i in $(find . -type d -name .settings | grep -v $reference_folder | grep -v
 do
 	process $i "org.eclipse.jdt.core.prefs"
 	process $i "org.eclipse.jdt.ui.prefs"
+	process $i "org.eclipse.pde.api.tools.prefs"
```
ruspl-afed, line 37: thanks!

ruspl-afed (approved):
LGTM

akurtakov (approved):
It's a huge patch! From what I've seen it should be ready to go. If smth is left if can always be finished and later PRs as moving to Java 21 is mandatory.

ruspl-afed:
> as moving to Java 21 is mandatory

What did I miss again @akurtakov ? 
I thought that Simpel members are free to use lower Java requirements than in SimRel.

merks:
I agree.  No one will manually review this much content!

No, it isn't required to switch everything to Java 21.  But the Lucene dependency generally requires it so definitely Mylyn as a whole cannot work with Java 21, though of course many parts could. 

Unless there is a somewhat compelling need to continue to work with Java 17, it seems easier to just do mass migration even though it's not required in general.

akurtakov:
> > as moving to Java 21 is mandatory
> 
> What did I miss again @akurtakov ? I thought that Simpel members are free to use lower Java requirements than in SimRel.

If there dependencies allow so. Lucene being kind of at the heart of Mylyn makes it not the case for Mylyn specifically IMO.

ruspl-afed:
Thank you for explanations @merks @akurtakov 

> I agree. No one will manually review this much content!

I tried very hard until GitHub started to collapse the contents of the file. Then I switched to check that modifications are done in the right files. Well, not enough, but better than review nothing.

Issue #658, opened 2024-12-29
Gitlab unit test failures


gnl42:
Hi @BeckerFrank, I'm seeing these Gitlab test failures against my multipass setup for `RestfulGitlabClientTest`:

| method | expected | actual |
| :-- | -- | ------------------------------------------------------------------------------------------------------------------------ |
| testGetUsers() | "private_profile": false, | "private_profile": true, |
| testgetConfiguration() |  | "ci_id_token_sub_claim_components": ["project_path", "ref_type",  "ref" ], |
| testGetGroups() | "default_branch_protection_defaults": {}, | "default_branch_protection_defaults": {"allowed_to_push": [{ "access_level": 40}], "allow_force_push": false, "allowed_to_merge": [ { "access_level": 40 } ],  "developer_can_initial_push": false }, |

I don't know if it is my multipass build, changes in the test data, the updated Jenkins instance, or ...

Thanks

BeckerFrank:
Hi @gnl42, I think we need to change the TestData!

In the tests I remove some known attributes where I know that the Value must not match (like creation timestamp) all other are in the compare. So If we have a failure I think we have a newer gitlab Instance with a changed behavior.
Yor setup should fav 17.4.0 and no longer 16.11.1.

If you want to make sure that I have missed to change the testdata, you can change line 117 of mylyn_setup.yml.

If not , I can verify this next year.  :smile:

gnl42:
I think I'll leave it up to you to decide which side of the compare has the correct value and make the appropriate changes

BeckerFrank:
Now the Testdata and the Version of the local instance are in sync.

PR #661, opened 2025-01-07
Update Gitlab Test Server to 17.7.0 #658

BeckerFrank:
in #658 we got errors because gitlab was updated to 17.4.0but not the testdata.
With this change we go to Version 17.7.0 and adjust the Testdata.
Now the tests also work when you change the localhost.test.server to an other value


gnl42 (approved):
LGTM.

PR #667, opened 2025-01-22
Include ECF dependencies not redistributed by the Platform in the site

merks:


merks:
FYI, this action is taken to be able to disable ECF's SimRel contribution.

image

I.e., preparation for 

https://www.eclipse.org/lists/cross-project-issues-dev/msg20026.html

I've tested that aggregation succeeds with these changes and with ECF disabled; I needed to fix tracecompass as well, but they don't actually have an ECF dependency just a dependency on a 3rd party bundle provided only by ECF's update site.

ruspl-afed (approved):
LGTM
Thank you for update @merks

PR #669, opened 2025-01-27
Update to ECF 3.15.5

merks:


ruspl-afed (approved):
Appreciate your effort to keep Mylyn up-to-date with ECF @merks 
I tried to understand what should we expect from ECF after announcement regarding leaving SimRel, but failed to interpret the message.
Please let me know if you need any assistance from my side.

merks:
This site has the same IU versions as the 3.15.4 site which is why I contribute now before m2 is done.   I'm also updating the platform where there are also no IU version changes:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2786

If I understand correctly, confusingly ECF wants to contribute for 2025-03 and then (maybe?) withdraw for 2025-06.  I don't see the real sense in that.   In any case, from a SimRel point of view, I want to verify that ECF can be disabled and that aggregation will be successful without it.   Currently I cannot update SimRel to 3.15.4 nor 3.15.5 anyway because of aggregation failures.  The dust should settle when mylyn and tracecompass have submitted updated contributions...

ruspl-afed:
> The dust should settle when mylyn and tracecompass have submitted updated contributions...

[here](https://github.com/eclipse-simrel/simrel.build/pull/704) is the mylyn part @merks

PR #678, opened 2025-02-25
Bump Develocity extension version

ribafish:
Update the Develocity extension version to 1.23.2, which brings support for Tycho test result discovery. This allows you to use build scans to view test failures and the exact reason for those, instead of getting the test report / searching through the console log, which can be especially useful on CI.


ruspl-afed (approved):
LGTM
Thank you for the improvement @ribafish

PR #692, opened 2025-03-18
Point to target as a file

akurtakov:
One less maven module to build.

### DIFF
--- a/mylyn.releng/org.eclipse.mylyn.parent/pom.xml
+++ b/mylyn.releng/org.eclipse.mylyn.parent/pom.xml
@@ -223,14 +223,7 @@
                 </environment>
               </environments>
               <target>
-                <artifact>
-<!-- revert back to ${project.parent.groupId} after Mylyn.Docs integration -->
-                  <groupId>org.eclipse.mylyn</groupId>
-<!-- revert back to ${project.parent.groupId}.target after Mylyn.Docs integration -->
-                  <artifactId>org.eclipse.mylyn.target</artifactId>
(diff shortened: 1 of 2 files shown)


mylyn.releng/org.eclipse.mylyn.parent/pom.xml
```diff
@@ -223,14 +223,7 @@
...
               <target>
-                <artifact>
-<!-- revert back to ${project.parent.groupId} after Mylyn.Docs integration -->
-                  <groupId>org.eclipse.mylyn</groupId>
-<!-- revert back to ${project.parent.groupId}.target after Mylyn.Docs integration -->
-                  <artifactId>org.eclipse.mylyn.target</artifactId>
-<!-- revert back to ${project.parent.version} after Mylyn.Docs integration -->
-                  <version>4.6.0-SNAPSHOT</version>
-                </artifact>
+                 <file>${maven.multiModuleProjectDirectory}/org.eclipse.mylyn-target/org.eclipse.mylyn.target.target</file>
```
ruspl-afed, line 226: Is this "the state of the art" @akurtakov ?

akurtakov: Let's say it's simplest way. I did it while trying to get dependabot to work on top dir as it doesn't work with pomless modules. Unfortunately, there are more pomless so giving up as dedicated time for that is over.

ruspl-afed: ok, thank you for the explanation

PR #707, opened 2025-04-22
Add SVGs to org.eclipse.mylyn Bundles

Michael5601:
This PR adds SVGs for all icons in the bundles `org.eclipse.mylyn.tasks.ui` and `org.eclipse.mylyn.wikitext.ui` except for the following as these are not available as SVG yet:

org.eclipse.mylyn.wikitext.ui/org.eclipse.mylyn.internal.wikitext.ui.util.resources/missing-image.svg

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.


ruspl-afed:
Thank you for your contribution @Michael5601 !

ruspl-afed:
@Michael5601 I think we need to update target for more recent platform build
```
[ERROR]   Software being installed: org.eclipse.mylyn.wikitext.ui 4.7.0.qualifier
[ERROR]   Missing requirement: org.eclipse.mylyn.wikitext.ui 4.7.0.qualifier requires 'eclipse.swt; (image.format=svg)' but it could not be found: See log for details
```

ruspl-afed:
Perhaps you can give me an advice re: build where it was merged to Platform.

Michael5601:
> Perhaps you can give me an advice re: build where it was merged to Platform.

The feature (and the new fragment org.eclipse.swt.svg) was merged in this [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638/files#diff-f8e5f9148a619fd6568f1ae371eb2c35a19c4672e4cfbdea7e462627b94323acR3).

merks:
It's a little strange, only if I explicitly specify `org.eclipse.swt.svg` in the target platform:

image

does it show up in the content:

image

Otherwise there is only this:

image

That doesn't really make sense to me. 

@laeubi @HannesWell 

 I anticipate that this is going to be a general problem downstream if requiring the feature doesn't automatically pull in the fragment as one would normally expect.

merks:
Hmm,  I can't reproduce the problem with a smaller sample target platform which works as expected...
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="New Target">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/eclipse/updates/4.36-I-builds"/>
			<unit id="org.eclipse.e4.rcp.feature.group"/>
		</location>
	</locations>
</target>
```

merks:
Oh wait, I didn't notice that the I build is hard coded!
```
			<repository location="https://download.eclipse.org/eclipse/updates/4.36-I-builds/I20250403-1800/"/>
```

ruspl-afed:
> Oh wait, I didn't notice that the I build is hard coded!

Yes, usually it points to M-build

merks:
Yes, that makes sense, but that's also why this PR can't move forward right now...

PR #708, opened 2025-04-22
Adapt functionality for using SVGs in org.eclipse.mylyn.tasks.ui

Michael5601:
This PR adapts functionality for using SVGs in the bundle `org.eclipse.mylyn.tasks.ui` and should be merged after [this](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/707).
As I don't have knowledge about potential side effects regarding this change I assumed it would be better to seperate these changes in this PR.

Please inform me if you see potential problems, so I can adjust the PR. :)


ruspl-afed (approved):
LGTM

PR #715, opened 2025-05-14
Remove Wizard SVGs

Michael5601:
This PR reverts the addition of SVGs for wizard icons of merged [PR](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/707/files). 

As stated by @HeikoKlare in this [comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2970#pullrequestreview-2839383590) the mechanism for creating SVGs for the wizard icons is not sufficient. This was my bad but some SVGs were already merged to Mylyn yesterday and thus need to be removed by this PR.

Please note that 5 SVGs where added yesterday which will be deleted by this PR. The reason why only 4 paths are changed is that for the icon `secur_role_wiz` the path was never changed to `.svg` and thus remains as `.gif`.


Michael5601:
@ruspl-afed @merks can you please check this small change?

ruspl-afed:
Ooops. TBH I had no time to check it in action. Could you please post here a screenshot how it looks with the content of this PR applied  @Michael5601 ?

Michael5601:
> Ooops. TBH I had no time to check it in action. Could you please post here a screenshot how it looks with the content of this PR applied @Michael5601 ?

Well it looks like the standard behaviour before the first change. In this PR I only reverted changes so the Wizard icons look like before (loaded with GIFs instead of SVGs). 

I have some problems starting an Eclipse Application with Mylyn right now so I sadly can't provide any screenshots.

ruspl-afed:
> so I sadly can't provide any screenshots.

I agree, this is sad. 
I'm going to give it a try, because I appreciate your effort to make  Mylyn better @Michael5601 
And please let me kindly ask you to have a look on it in upcoming M3,

ruspl-afed:
> I will try to fix my setup for mylyn as soon as I find some time, then I can provide the screenshots :)

Definitely, it is good to have working Mylyn setup. However I hope you can also try to grab any M3 package with Mylyn tomorrow

PR #673, opened 2025-01-28
Fixed wrong project-id referenced in the Jenkinsfile. #664

ribafish:
Fixed wrong project-id referenced in the Jenkinsfile. The project-id needs to be the id from [project.eclipse.org](http://project.eclipse.org/): https://projects.eclipse.org/projects/tools.mylyn, instead of http://ci.eclipse.org/. I will also update the [documentation](https://gitlab.eclipse.org/eclipsefdn/it/releng/develocity/develocity-documentation#ci-integration).

Related to issue #664.

PR #665, opened 2025-01-13
Develocity configuration

ribafish:
This PR will enable you to publish Build Scans to [develocity-staging.eclipse.org](https://develocity-staging.eclipse.org) as requested [here](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5409).

### Description
This PR publishes a build scan for every CI build and for every local build from an authenticated Eclipse committer. The build will not fail if publishing fails. Local and remote caching was left disabled on this PR by design so that the build is not affected by this change.

The build scans of the Eclipse Mylyn project are published to the Develocity instance at [develocity-staging.eclipse.org](https://develocity-staging.eclipse.org), hosted by the Eclipse Foundation and run in partnership between the Eclipse and Gradle. This Develocity instance has all features and extensions enabled and is freely available for use by the Eclipse Mylyn project and all other Eclipse projects.

On this Develocity instance, Eclipse Mylyn will have access not only to all of the published build scans but also to other aggregate data features such as:

- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

This will also enable you to (optionally) use build time optimization features, such as (remote) build caching and Predictive Test Selection. 

More information can be read in the [Eclipse announcement](https://www.eclipse.org/lists/cross-project-issues-dev/msg20000.html). This is the first part of Develocity integration issue no. #664. The second part, if you opt for it, would be to enable caching for the project.

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.

Locally, you can authenticate by running `mvn develocity:provision-access-key`. More info available [here](https://docs.gradle.com/develocity/maven-extension/current/#automated_access_key_provisioning)

PR #718, opened 2025-05-19
Develocity update with tycho caching

ribafish:
Work done:
- Update the Develocity extension to version 2.0, which brings Tycho tests caching support
- Configure additional inputs to be considered when caching Tycho tests
    - These are the `config.ini` and `.eclipse.keyring` files, the `testdata` directory and the `test-fixture.xsl` and `test-suite.xsl` files in the `org.eclipse.mylyn.tests` project. 
- Enable caching locally
- Enabled caching on CI (Jenkins)
- Remove the `maven.test.failure.ignore=true` and the deprecated `maven.test.error.ignore=true` flags set for CI builds. 
    - I checked the CI test history and found that all tests are being reported as passing, so I thought it safe to do so. This is important, as we don't want to cache failing tests, which is caused by using this flag. 
    - An alternative that works well with caching would be to use the `--fail-at-end` flag instead, to run all the tests (except for the ones downstream from the module containing the test failure), but still fail the build at the end.
 
Since I'm not an expert in how Mylyn build and tests are configured, please let me know if:
- There are other files that are used for tests
- Would like to discuss the test failure ignore strategy further. 
- Have any other questions, topics, or concerns that you'd like to clarify or discuss

Please test this well, but in general, I see a very good opportunity for building time improvement here, as the Mylyn project is built nightly, often without any changes. For those cases, the build time should be cut down by roughly 18minutes, which would be a substantial improvement, since the average build time in the last 28 days for CI is [24m 51s](https://develocity-staging.eclipse.org/scans/trends?search.buildToolType=maven&search.relativeStartTime=P28D&search.rootProjectNames=Mylyn%20Root&search.tags=ci&search.timeZoneId=Europe%2FLjubljana&trends.section=buildTime) 😄


Jenkinsfile
```diff
@@ -89,8 +89,6 @@ MAVEN_PROFILES=${env.MAVEN_PROFILES}
                                 -B \
                                 $MAVEN_PROFILES \
                                 -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-                                -Dmaven.test.failure.ignore=true \
```
ruspl-afed, line 92: Is it really related with the rest of changes?

ruspl-afed (changes requested):
I would prefer to have one commit for related changes, now we have 6 commits.

Jenkinsfile
ribafish: Yes, this is important, as we don't want to cache failing tests, which is caused by using this flag. I can also open a separate PR just with this, but it is related. 

An alternative that works well with caching would be to use the --fail-at-end flag instead, to run all the tests (except for the ones downstream from the module containing the test failure), but still fail the build at the end. 

Note: I'm happy to squash the commits, but I suggest we resolve any open questions beforehand.

ruspl-afed: ok, let's keep it here and give it a try

ruspl-afed:
@ribafish unfortunately, I had to disable devolocity completely, otherwise I would not have been able to create an M3 build

ribafish:
Hi @ruspl-afed, from what I [see on the CI](https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/474/console), it looks like the issue at hand wasn't really the changes done with this PR (caching, or test failure ignore), but rather the vault configuration, which was done as part of the [initial connecting PR](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/665) (and the [fix PR](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/673)), as we can see this error at the end:

```
com.bettercloud.vault.VaultException: Vault responded with HTTP status code: 503
```  

I know there were recently some Eclipse Foundation infrastructure issues, are you perhaps aware if this might be connected, or was it maybe just a transitive issue with the vault availability? Would you mind reverting the changes you did in commits [e27a1b](https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/e27a1b50bd5a37f57c92be6a473585dd70d60f2a) and [dd283e](https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/dd283e755d534dc2838ae1cff0c1fa0caf87d300) to test whether the issue still persists (on a branch if necessary) and if it is, could you open a [helpdesk](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues) ticket?

ruspl-afed:
@ribafish I'm not saying that your changes were breaking. And yes, it looks more like an infra issue.

Please consider the fact that we are in RC cycle now and we need to have green master without surprises.
You are welcome to work on this in a dedicated branch, open helpdesk tickets, and so on.
But all the experiments in master may happen only after RC2 contribution to SimRel.

PR #734, opened 2025-06-20
Fix build

akurtakov:
4.36-I-builds are gone now thus switch to 4.37/2025-09


ruspl-afed (approved):
Thanks @akurtakov !
I plan to bump versions soon

Issue #534, opened 2024-05-21
multipass update 2


gnl42:
Update Jenkins version, some comments

BeckerFrank:
@gnl42 I have updated https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/652 so that the following url (hopefully accessible from the Internet) temporarily displays the list of services hosted on my local network.

So please try whether https://mylyn.frank-becker.de/mylyn_idx/ displays a list of Bugzilla, Jenkins and Gitlab. Also please try if you can access all the services listed. I think we may have a problem with the Gitlab URL.

gnl42:
@BeckerFrank The server is not responding. I'll try again tomorrow

BeckerFrank:
> @BeckerFrank The server is not responding. I'll try again tomorrow

Maybe this is because my Multipass instant is on a Macbook and was in sleep mode.

gnl42:
Still no response

BeckerFrank:
I think we can close this ticket.

PR #754, opened 2025-07-19
[#534] save the multipass setup for the ci server

BeckerFrank:
This is the actual multipass configuration for the ci server on my Mac mini M4


ruspl-afed (approved):
LGTM in general
The only questionable point for me is why GitLab tests failed for one of the pipe. Restarted it.

Issue #672, opened 2025-01-28
Validdations doesnt't take credentials parameters into considation.


emaayan:
when creating a new Task repo for github, validations button doesn't take wrong credentials into consideration.
meaning you can input anything into the user password, and it would still pass.

BeckerFrank:
For some reason the request https://api.github.com/repos/<repo path>/issues?state=open&per_page=100&page=1 with Authorization request header and value token github_pat_.... returns a 401 status wit an invalide Authorization value.

But with Basic authorization I always get a 200 response.

Thougths?

BeckerFrank:
Because gitlab disabled user and password for Oauth2 we can only work with tokens.

PR #755, opened 2025-07-20
Gitlab Connector: remove user and password authorization #672

BeckerFrank:
Gitlab authorization over oauth no longer accepts user and password.
So we must change the repository settings to always use the token and remove unused code.


ruspl-afed (approved):
LGTM

Issue #761, opened 2025-07-23
Continuous Integration / Test Results (push) - 29 errors, 12 fail


ruspl-afed:
We ignored test failures and errors for a long time, that is not good.

All the errors and failures should be either fixed or ignored with the link to corresponding issue.

BeckerFrank:
I can see if I can find the source of the problems.
Maybe some of them are related to line termination issues between Unix and Windows, because we only see the errors in Windows tests.

ruspl-afed:
If you can extract concrete tasks from this investigation this would be great @BeckerFrank 

regarding test env we can either parametrize reused Platform workflow or create our own.

ruspl-afed:
extreme progress @BeckerFrank , thank you!

merks:
Awesome 🤩

ruspl-afed:
Thank you so much for your effort @BeckerFrank !

I wonder why it reports failure and do not want to became green

BeckerFrank:
The junit annotation @Ignore did not work for junit3 style tests (extension of class TestCase).
Should we create a ticket and convert everything to junit4 / junit5 style?

PR #765, opened 2025-07-26
fix for 6 errors and 1 warning(#761)

BeckerFrank:


ruspl-afed (dismissed):
Good to go, with minor notes that can be fixed later

PR #766, opened 2025-07-26
[#761] Continuous Integration / Test Results (push) failure

ruspl-afed:
* stop ignoring test errors and failures


ruspl-afed:
Let's see what is still failing

ruspl-afed:
Hmm, now it say even more strange things
image

ruspl-afed:
TBH I do not understand what is going on and how many flaky tests we have

BeckerFrank:
> TBH I do not understand what is going on and how many flaky tests we have

Maybe we have a problem when at the workload of the test instance is to high.

gnl42:
> > TBH I do not understand what is going on and how many flaky tests we have
> 
> 
> 
> Maybe we have a problem when at the workload of the test instance is to high.

Time based tests are never a good idea

ruspl-afed:
> Time based tests are never a good idea

Agree. Feel free to improve this situation,

BeckerFrank:
> > > TBH I do not understand what is going on and how many flaky tests we have
> > 
> > 
> > Maybe we have a problem when at the workload of the test instance is to high.
> 
> Time based tests are never a good idea

Yes but sometimes we need Time based tests like in MonitorUserActivityJobTest oe is there an other option for this kind of tests?

Issue #756, opened 2025-07-22
Reuse platform CI


ruspl-afed:
Platform has notable improvements in CI process and we may reuse a lot from there
* unit-tests for different platforms
* unit-test result publishing
* more

PR #777, opened 2025-08-09
[#756] Reuse platform CI

ruspl-afed:
* run sync-code-style


gnl42:
@ruspl-afed There should be a lot more files changed.

Looking at the cleanup settings for the org.eclipse.mylyn.commons.core project I don't see any changes that match the cleaned up code you worked on.
E.g.:
image

ruspl-afed:
I suspect `tycho-cleancode-plugin` is configured by https://github.com/eclipse-mylyn/org.eclipse.mylyn/blob/f5492b708e425fe1641b07e2b040f3c1478455fc/mylyn.releng/org.eclipse.mylyn.parent/pom.xml#L372
However it is better to ask expert
@laeubi how `tycho-cleancode-plugin` interacts with cleanup project settings?

laeubi:
@ruspl-afed yes currently it uses the settings from the `<cleanUpProfile>` exclusively (as a mean for a global policy).
For the jdt compiler plugin, we have a mean of enable a setting called [useProjectSettings](https://tycho.eclipseprojects.io/doc/latest/tycho-compiler-plugin/compile-mojo.html#useProjectSettings), I could imagine such parameter would be useful here as well, please don't hesitate to suggest a PR for adding such thing to Tycho (it could be default on for Tycho 6 but should be default off for Tycho 5 backport then).

gnl42:
So the plan is to have two "competing" cleanup processes?

ruspl-afed:
> So the plan is to have two "competing" cleanup processes?

The plan is to automate cleanup process. I do not see any competition here. 
Yes, currently we have configuration divergence, but this is something to fix first manually (improving plugin configuration) and then with improving Tycho itself

PR #778, opened 2025-08-09
[#756] Reuse platform CI

ruspl-afed:
* use Mylyn release repo for baseline
* add latest Mylyn nightly to regular target to make it resolvable for individual bundles


ruspl-afed:
@laeubi may be you know better way to make target resolvable when running for single bundle: my idea here is to add latest nightly. Not sure if this could harm anything.

laeubi:
> @laeubi may be you know better way to make target resolvable when running for single bundle: my idea here is to add latest nightly. Not sure if this could harm anything.

Yes platform uses the same approach with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/bb531ea415d5aa7dc575582b47d1425334f4799f/eclipse-platform-parent/pom.xml#L796-L820

Issue #575, opened 2024-07-14
Remove guava from wikitext, part 3


gnl42:
Remaining  guava code left over after #521 and #568

gnl42:
@BeckerFrank  I'll take another look at what's left.

PR #802, opened 2025-08-11
Remove last vestiges of guava #575

gnl42:
Remove last vestiges of guava

### DIFF
--- a/org.eclipse.mylyn-target/org.eclipse.mylyn.baseline.target
+++ b/org.eclipse.mylyn-target/org.eclipse.mylyn.baseline.target
@@ -118,7 +118,6 @@
 			<unit id="org.apache.tika.core" version="0.0.0"/>
 			<unit id="com.github.ben-manes.caffeine" version="0.0.0"/>
 			<unit id="com.google.gson" version="0.0.0"/>
-			<unit id="com.google.guava" version="0.0.0"/>
 			<unit id="org.hamcrest" version="0.0.0"/>
 			<unit id="org.jsoup" version="0.0.0"/>
 			<unit id="org.junit" version="0.0.0"/>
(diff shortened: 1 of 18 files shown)


org.eclipse.mylyn-target/org.eclipse.mylyn.baseline.target
```diff
@@ -118,7 +118,6 @@
 			<unit id="org.apache.tika.core" version="0.0.0"/>
 			<unit id="com.github.ben-manes.caffeine" version="0.0.0"/>
 			<unit id="com.google.gson" version="0.0.0"/>
-			<unit id="com.google.guava" version="0.0.0"/>
```
ruspl-afed, line 121: this should be done after this release, since we cannot change the past

ruspl-afed (commented):
I'm not fluent in regexps to judge the related changes, but in general it looks good.
The only point is baseline change: I believe it should be reverted @gnl42

org.eclipse.mylyn-target/org.eclipse.mylyn.baseline.target
ruspl-afed: Here the "baseline" is the state of the target of the previous release, typically used for API control and things like that,
This is why I say that "we cannot change the past", because the previous release was in the past.
After every release we normally should replace "baseline target" with "target".

gnl42: Gotcha. It has been reverted

gnl42:
@ruspl-afed Any idea what is going on?
`Error:  Failed to execute goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check (default-cli) on project org.eclipse.mylyn-aggregator: Execution default-cli of goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check failed: Received an error response from ClearlyDefined. -> [Help 1]`

ruspl-afed:
> Received an error response from ClearlyDefined

This is the 3rd party infra failure @gnl42 
since we had no new dependencies it was safe to merge

PR #864, opened 2025-08-31
add getter for field message #831

BeckerFrank:
This change is intended to prevent the cleanup from removing the unused Message field.


gnl42 (approved):
LGTM

Issue #868, opened 2025-09-03
Prepare releng for 4.9.0 development cycle


ruspl-afed:
- [x] Increment manifests versions for plugins and features
- [x] Increment version for poms
- [x] update baseline target

merks:
How dare you get sick!  👿   Just kidding, I hope you feel better.  👼

PR #872, opened 2025-09-14
Use 4.38-I-builds because all 4.37-I-builds are deleted

merks:
https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/868


merks:
@gnl42 

It looks like some service used to test bugzilla is unavailable:

image

I started another build to see if it's just a glitch.

In any case, the problem is not related to this PR.

ruspl-afed:
Thank you for working on this @merks 
Do you need M0 or something similar from Mylyn?

merks:
No aggregation is running fine at the moment.

PR #876, opened 2025-09-20
[#868] Prepare releng for 4.9.0 development cycle

ruspl-afed:
* Increment manifests versions for plugins and features
* Increment version for poms
* update baseline


gnl42 (approved):
Is there a reason why we have various imports with varied versions: 4.2, 4.3.0, 4.4.0, 4.9.0

ruspl-afed:
No idea what is so wrong, but it started from switching to 4.38 integration stream

Issue #894, opened 2025-09-26
CoreUtil.getRuntimeVersion() fails to parse Java version properly


gnl42:
CoreUtil.getRuntimeVersion() is unable to convert "21.0.8+9-Ubuntu-0ubuntu125.04.1" to a org.osgi.framework.Version which is used to check for broken certificate using CommonTestUtil.isCertificateAuthBroken()

A certificate is "considered" broken if java version <= "1.6.0.25" so is the code still needed?

Similar to  b6cd39edb292779274302df262660523c5a665a5

merks:
Consider to use Runtime's feature(), interim(), update(), patch() instead.  And yes, it can't be < your current BREE.

PR #895, opened 2025-09-27
CoreUtil.getRuntimeVersion() fails to parse Java version properly #894

gnl42:
Replace custom java version parser with native version, if needed.

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/894

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.core/src/org/eclipse/mylyn/commons/core/CoreUtil.java
+++ b/mylyn.commons/org.eclipse.mylyn.commons.core/src/org/eclipse/mylyn/commons/core/CoreUtil.java
@@ -177,7 +177,7 @@ public class CoreUtil {
 	}
 
 	/**
-	 * @since 4.9
+	 * @since 4.3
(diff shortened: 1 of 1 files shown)


merks (approved):
That's much simpler.

mylyn.commons/org.eclipse.mylyn.commons.core/src/org/eclipse/mylyn/commons/core/CoreUtil.java
```diff
@@ -170,61 +170,33 @@ private static void defaultSystemProperty(String key, String defaultValue) {
...
 	 */
 	public static Version getRuntimeVersion() {
-		Version result = parseRuntimeVersion(Runtime.version().toString());
-		return result;
+		return convertToOsgiVersion(Runtime.version());
 	}
 
 	/**
-	 * @since 4.3
+	 * @since 4.9
```
ruspl-afed, line 180: We do not change signature, right?
So we should keep "since" value as well.

Issue #51, opened 2023-02-03
Migrate to JUnit 5


gnl42:

BeckerFrank:
What is the state of this issue?
Are you working on this?

gnl42:
I put it on hold for now. There is something hopefully in 5.12 that I think will be a better fit

https://github.com/junit-team/junit5/issues/878#issuecomment-2349215835

gnl42:
@BeckerFrank I'm want to re-enable the tests that use "http://mylyn.org", and there are some entry points that don't appear in your multipass implementation.

Most I can simply use your M4 for (with some minor url changes) but I was wondering if there are any entry points that would work for:

- CommonHttpClientTest::testHttpAuthenticationTypeHttp (and friends)
- ServiceMessageManagerTest:::testRetrievingMessage

For CommonHttpClientTest::testGetRequest() (end friends) I figure I can use https://mylyn.frank-becker.de/mylyn_idx/service instead of http://mylyn.org as a low cost entry point.

Thanks

gnl42:
junit4 is gone

PR #896, opened 2025-09-29
junit bugzilla connector #51

gnl42:
Convert Bugzilla Connector to using JUnit 5 for unit tests.

The JUnit 4 dynamic `TestSuite` is not available for JUnit 5 so created a "framework" to provide similar functionality, using Extensions, `@SelectClasses` coupled with `@BeforeEach` runtime checks.

Added a new module, `org.eclipse.mylyn.commons.sdk.util.junit5`, that adds:

1. `@EnabledIfCI`: Uses `CommonTestUtil.runNonCIServerTestsOnly()` to enable or disabled test(s)
2. `@MylynTestSetup`: Sets up JUnit 5 extensions to register:
    - `TestMethodWrapper`: 
        - prints test name  when started
        - does a thread dump if a test is aborted due to `@Timeout`
    - `TestResultLogger`: reports test result status
3. `SuiteFailureSummaryListener`: when registered using `META-INF/services/org.junit.platform.launcher.TestExecutionListener` prints a summary of failed/disabled/skipped tests in a `@Suite`

`SimpleTester` provides a simple tester of the framework.

### DIFF
--- a/mylyn.tasks/connectors/bugzilla/org.eclipse.mylyn.bugzilla.tests/src/org/eclipse/mylyn/bugzilla/tests/AbstractBugzillaFixtureTest.java
+++ b/mylyn.tasks/connectors/bugzilla/org.eclipse.mylyn.bugzilla.tests/src/org/eclipse/mylyn/bugzilla/tests/AbstractBugzillaFixtureTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 george
+ * Copyright (c) 2025 George Lindholm
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at


merks (commented):
I'm really not qualified to review this.

BeckerFrank:
We have some test classes with incorrect parent classes.
For example, BugzillaVersionTest was previously a subclass from TestCase and is now derived from AbstractBugzillaFixtureTest.
As a result, the version test is not only executed once, but is called for each fixture.

ruspl-afed (approved):
Looks good to try it in action.
Thank you @gnl42

BeckerFrank:
> Looks good to try it in action. Thank you @gnl42

Why did you ignore  my comment

> We have some test classes with incorrect parent classes. For example, BugzillaVersionTest was previously a subclass from TestCase and is now derived from AbstractBugzillaFixtureTest. **As a result, the version test is not only executed once, but is called for each fixture.**

Issue #897, opened 2025-10-08
Publish feature with Mylyn test to improve CI checks


ruspl-afed:
We have a number of failures in CI for test bundle code audits, because test bundles are not published and cannot be referenced for target resolution

The suggestion is to publish a feature with test bundles.

PR #898, opened 2025-10-08
[#897] Publish feature with Mylyn test to improve CI checks

ruspl-afed:
* publish `org.eclipse.mylyn.tests.feature`


ruspl-afed:
BTW, very interesting OS-specific timing @gnl42 
image

Issue #934, opened 2025-10-25
add junit5 to target platform


BeckerFrank:
Jenkins Pipeline for Pull Request fail with
```
[ERROR] Cannot resolve dependencies of project org.eclipse.mylyn:org.eclipse.mylyn.commons.sdk.util.junit5:eclipse-plugin:4.9.0-SNAPSHOT
[ERROR]  with context {osgi.os=macosx, osgi.ws=cocoa, org.eclipse.update.install.features=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
[ERROR]   Software being installed: org.eclipse.mylyn.commons.sdk.util.junit5 4.9.0.qualifier
[ERROR]   Missing requirement: org.eclipse.mylyn.commons.sdk.util.junit5 4.9.0.qualifier requires 'java.package; org.junit.jupiter.api [5.13.1,6.0.0)' but it could not be found: See log for details
```

so we should add the junit 5.14 to the target platform.

PR #935, opened 2025-10-28
[#934] add junit5 to target platform

ruspl-afed:
* add JUnit 5 with required libraries


ruspl-afed:
@merks may be you can have a look at this. 
Some time ago we added JUnit `[5.13.1,6.0.0)` constraint to a few test bundles.
But then, since we use Orbit from dynamic 2025-12 URL, we started to receive JUnit6 and fail.
My solution is to relax JUnit constraints to `[5.13.1,7.0.0)`
[An alternative](https://github.com/eclipse-mylyn/org.eclipse.mylyn/commit/1f7015ed211666e19c65b60a575a46f922befa7d) from @BeckerFrank is to add JUnit 5 from Maven to `.target` that I like a bit less, because I want to keep "3rd party currency" with SimRel

merks (commented):
Note that Tycho and the Platform support version now ranges so you might want to make this more flexible with `[5.0.0,6.0.0)` and `[1.0.0,2.0.0)` so that you don't have to change this when there are new versions in Orbit...

PR #957, opened 2025-11-20
Use org.apache.tika 3.x

merks:
- The 2.x version has a CVE and is removed from Orbit.


merks:
I'll make an executive decision to merge this.  The build would be broken without it given the 2.x version is removed from Orbit's milestone.

PR #988, opened 2025-12-12
Fix the target to build currently

merks:


ruspl-afed (approved):
LGTM
Thank you @merks

Issue #1031, opened 2026-03-20
TaskEditor: Graphic is disposed


gnl42:
Seeing this from the JiraConnector when focus is lost or gained:
```
!ENTRY org.eclipse.ui 4 0 2026-03-19 16:36:42.240
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.swt.SWTException: Graphic is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4950)
	at org.eclipse.swt.SWT.error(SWT.java:4865)
	at org.eclipse.swt.SWT.error(SWT.java:4836)
	at org.eclipse.swt.graphics.Image.getImageData(Image.java:1348)
	at org.eclipse.mylyn.tasks.ui.editors.TaskEditor$7.drawCompositeImage(TaskEditor.java:787)
	at org.eclipse.jface.resource.CompositeImageDescriptor.getImageData(CompositeImageDescriptor.java:376)
```

gnl42:
Here is the allocation stack, in case it might make sense to someone:
(Interestingly enough it happened when I submitted this comment)
```
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:191)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:124)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:735)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:382)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:331)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:309)
	at org.eclipse.mylyn.tasks.ui.editors.TaskEditor.setHeaderImage(TaskEditor.java:797)
	at org.eclipse.mylyn.tasks.ui.editors.TaskEditor.updateHeaderLabel(TaskEditor.java:1142)
	at org.eclipse.mylyn.tasks.ui.editors.TaskEditor$1.controlResized(TaskEditor.java:233)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:277)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4349)
[...]

PR #1032, opened 2026-03-20
TaskEditor: Graphic is disposed #1031

gnl42:
Don't try to use image if it is disposed

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1031


ruspl-afed (approved):
LGTM

PR #1042, opened 2026-04-10
DatePicker UI #1039

gnl42:
### Got tired of seeing the DatePicker (and friends) truncated in Windows:
image
as well as in the JiraConnector in Windows:
image

Linux/MacOs was better but also had some issues:
image

With these changes it looks like:
image
and
image
and
image

### The DatePicker popup was cleaned up a bit as well from:
image
to
image

**For some reason that I don't understand, the Mylyn-Runtime.launch version differs slightly in appearance from when Mylyn is installed in an Eclipse instance.**

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.ui/src/org/eclipse/mylyn/commons/ui/FontUtils.java
+++ b/mylyn.commons/org.eclipse.mylyn.commons.ui/src/org/eclipse/mylyn/commons/ui/FontUtils.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+ * Copyright (c) 2026 George Lindholm
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html.
+ *
+ * Contributors:
+ *      See git history
+ *******************************************************************************/
+
+package org.eclipse.mylyn.commons.ui;
+
+import org.eclipse.swt.graphics.FontMetrics;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.widgets.Control;
+
+/**
+ * Some utility methods for working with fonts.
+ * 
+ * @since 4.11
+ */
+public class FontUtils {
+	private static final Point POINT_ZERO = new Point(0, 0);
+
+	/**
+	 * Returns the font metrics for the font of the given control.
+	 *
+	 * @param control
+	 * @return
+	 */
+	public static FontMetrics getFontMetrics(final Control control) {
+		final GC gc = new GC(control);
+		try {
+			return gc.getFontMetrics();
+		} finally {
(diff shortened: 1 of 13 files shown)


mylyn.commons/org.eclipse.mylyn.commons.ui/src/org/eclipse/mylyn/commons/ui/FontUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.widgets.Control;
+
+/**
+ * Some utility methods for working with fonts.
+ * 
+ * @since 4.11
+ */
+public class FontUtils {
```
ruspl-afed, line 24: Although added methods looks useful, I would avoid introducing new API during this change just to not mix two different discussions.

gnl42: Deleting it and then re-creating it doesn't sound very appealing.

Any other options?

gnl42: @ruspl-afed what would you like me to do?

ruspl-afed: Sorry, I missed your reply here. However, I do not understand it.
My point was to not introduce new API here, and focus on the UI itself.
And this doesn't mean deletion and re-creation, it means moving new type to an internal package for further API-related discussion.

gnl42: I think it should be ok now

ruspl-afed: Now we have two `FontUtils` types. I suppose we need inly one, located at `internal.commons.ui`

gnl42: Sigh, the refactor move only committed the new location, leaving the original location (and references) in place.

I've been seeing this recently, I do a commit only to have some of the changed  files suddenly appear as unstaged afterwards.
Both in Windows and Ubuntu

ruspl-afed: The only thing to do here is to squash to single commit

gnl42: Done

gnl42:
Wondering if we should treat fixture discovery failure as a "disabled/skipped" test?

ruspl-afed (approved):
LGTM

Issue #1044, opened 2026-04-12
Eclipse 4.40.0-M1 breaks workspace


gnl42:
Don't know what changed with the Eclipse M1 but the workspace is having problems:

- The package org.w3c.dom is accessible from more than one module: <unnamed>, java.xml
- The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml
- The package javax.xml.transform is accessible from more than one module: <unnamed>, java.xml
- One or more cycles were detected in the build path of project 'org.eclipse.mylyn.wikitext.tasks.ui'. The paths towards the cycle and cycle are:
org.eclipse.mylyn.wikitext.tasks.ui->{org.eclipse.mylyn.wikitext.tracwiki, org.eclipse.mylyn.wikitext.tracwiki.tests}
org.eclipse.mylyn.wikitext.tasks.ui->{org.eclipse.mylyn.wikitext.tracwiki, org.eclipse.mylyn.wikitext.tracwiki.tests}
org.eclipse.mylyn.wikitext.tasks.ui, org.eclipse.mylyn.wikitext.tracwiki.ui->{org.eclipse.mylyn.wikitext.tracwiki, org.eclipse.mylyn.wikitext.tracwiki.tests}

Interestingly enough, the `mvn verify` runs with no errors.

This is a fresh vanilla mylyn oomph  install

merks:
Add this to the eclipse.ini
```
-Dpde.addTransitiveDependenciesWithForbiddenAccess=false
```
I updated my local install and that helped eliminate this problem.

@HannesWell  @laeubi 

This is clearly a very serious issue in complex workspaces that are very far along the food chain with all kinds of upstream dependencies dragging in all kinds of stuff beyond the project's own control.  In this case, is there really anything the project can do about the new problems from the transitively large classpath?  I.e., it appears the classpath computation has created a cycle (maybe the project's own fault) and has made the same package visible from more than one bundle.

(I'm scared to update the BIRT IDE now.)

merks:
For what it's worth, my BIRT IDE updated fine...  🥳

HannesWell:
> One or more cycles were detected in the build path of project

Thanks for the heads up. 
That's an interesting new error, but I cannot tell from my head why this could be caused by the issue Ed is thinking of. 
If there is a cycle now, there should have been one before, it's now just direct and not indirect.

But I can try to reproduce it with the oomph set-up in the next days.

[...]

PR #1046, opened 2026-04-14
eclipse 4.40 #1044

gnl42:
Making the workspace workable again after Eclipse 4.40-M1 came out, see #1044 

I don't know what changed to break httpcomponents.{client,core}, but making them required instead of imported fixed running 
org.eclipse.mylyn.commons.repositories.http.tests

All the tests run under Eclipse (several fail), but `mvn verify` is passes


ruspl-afed:
@merks do you have concerns regarding switching from `Import-Package` back to `Require-Bundle` for `org.apache.httpcomponents.*` libs from SimRel perspective?
I would merge this to let @gnl42 have progress with other tasks, and create an issue to align Apache HTTP dependencies

mylyn.commons/org.eclipse.mylyn.commons.repositories.http.tests/META-INF/MANIFEST.MF
```diff
@@ -5,7 +5,9 @@ Bundle-Name: Mylyn Commons Tests
 Bundle-SymbolicName: org.eclipse.mylyn.commons.repositories.http.tests;singleton:=true
 Bundle-Version: 4.11.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
-Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
+Require-Bundle: org.apache.httpcomponents.httpclient;bundle-version="4.5.14",
```
merks, line 8: I'm confused that this is here, but removed from the target below.

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
```diff
@@ -58,7 +58,6 @@
 		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
 			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20221123021534/repository/"/>
 			<unit id="jakarta.servlet" version="0.0.0"/>
-			<unit id="org.apache.httpcomponents.httpclient" version="0.0.0"/>
```
merks, line 61: I'm a bit confused this is removed but used above.

merks (commented):
I'm not religious about bundle require versus package import.  From a SimRel point of view I only care that it's possible to update 3rd party bundles without needing updating project content which is mostly an issue of including the 3rd party bundles in features because that restrict to exactly and only that one version.

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
gnl42: There is something very weird going on with library resolution and the target file.

If I kept the target entry, the unit test fails to run under eclipse with:
image

Something similar seems to apply to the codec library, it was no longer being resolved so I tried adding it to target, and it still wasn't resolving, so I removed all references to it.

merks: There are multiple versions of codec with 1.20 being newer.

image

Using it for base64 support is probably overkill.

The problems look more like you need to do "Select Required"...

When I try to launch indeed it complains but when I look at the launch configuration it's clearly not going to work properly:

image

@HannesWell @laeubi

This seems a strange state for this to be in initially.

Even after Select Required it still has issues:

image

And while the necessary bundle is present, it's not been selected:

image

Selecting that requires more things:

image

But Select Required does not add those.

image

I have to deselect this to make all the validation errors go away:

image

That's quite painful and seems significantly PDE-buggy to me.

Once the right magic is applied, the test does run without the changes to the dependencies.

image

It seems less than ideal to make changes that effectively only necessary to work around what appear to be PDE bugs.    E.g., how can adding a bundle suddenly make a package not be resolved?

image

In this state, the launch does work and the tests do pass.

Something is serious amiss here with PDE determination of dependencies and it's ability to compute the required ones, apparently complicated by the use of package imports versus bundle requirements.

merks: I think an underlying problem that is not really a PDE problem is that the bundles themselves are crappy:

So we see that target platform state that the old codec bundle allows to substitute the packages of the newer (different BSN) code bundle but that has a higher version and httpclient doesn't like that:

image

And I'll bet if we debugged PDE we'd find it didn't add httpclient because it has unresolved requirements, though not adding it results in an even large set of unresolved requirements.

Ideally "we/Mylyn" would use httpclient5 or just the JDK would be much much better.

ruspl-afed: Thank you for this excellent investigation @merks !
I agree with your conclusion to retire http < 5 and plan to create the corresponding issue for Mylyn

gnl42: My vote is for native java

ruspl-afed:
@gnl42 please align changes in `MANIFEST.MF`, `.target` and `category.xml` manifests.
Ideally removing of "codec" and changes around "http" should go as a separate PRs.
Changes in `.setup` looks good but not related to the dependency changes discussed above,
Also, it would be great to have one commit per PR, perhaps I'll enforce this via pipeline configuration

gnl42:
All the changes are part of the same issue. It doesn't make sense to split them out.

Short of doing a forced push, I don't know of any way to combining changes found during the PR build after the initial commit

ruspl-afed:
> All the changes are part of the same issue. It doesn't make sense to split them out.

Even if we assume that this is the case, what about removal `org.apache.httpcomponents.httpclient` from `.target`, is it still required?

gnl42:
> > All the changes are part of the same issue. It doesn't make sense to split them out.
> 
> Even if we assume that this is the case, what about removal `org.apache.httpcomponents.httpclient` from `.target`, is it still required?

Yes, see my post https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/1046#discussion_r3084468223

ruspl-afed:
> Short of doing a forced push, I don't know of any way to combining changes found during the PR build after the initial commit

There is nothing bad in a force push to a dedicated branch, since for a focused change all the intermediate steps to find a solution are either not so important or may be documented in a commit message. From change management perspective, it is more valuable to see the `main` branch as a sequence of consistent (green & runnable) states.

ruspl-afed:
> Yes, see my post [#1046 (comment)](https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/1046#discussion_r3084468223)

@merks please complete your review (and please resolve conversations above if you are OK with this)

gnl42:
> > Short of doing a forced push, I don't know of any way to combining changes found during the PR build after the initial commit
> 
> There is nothing bad in a force push to a dedicated branch, since for a focused change all the intermediate steps to find a solution are either not so important or may be documented in a commit message. From change management perspective, it is more valuable to see the `main` branch as a sequence of consistent (green & runnable) states.

so something like?:
1. git pull
2. git checkout -b work-needed origin/main
3. Some work
3. git commit -m 'some work'
4. Some more work
5. git commit -m 'something else fixed'
6. git checkout -b issue-work-needed origin/main
7. git merge --squash work-needed
8. git commit -m 'Fixed bad bug'
9. git push -u origin issue-work-needed
10. Feedback on PR
11. git checkout work-needed
12. Address PR comments
13. git commit -m 'address PR comment'
14. git rebase origin/main
15. git branch -D issue-work-needed
16. git checkout -b issue-work-needed origin/main
17. git merge --squash work-needed
18. git commit -m 'Fixed bad bug'
19. git push --force-with-lease -u origin issue-work-needed
repeat 11–19 as needed

ruspl-afed:
> so something like?:

Personally I'm using EGit UI, and the concrete sequence of steps may be different, but the idea is to always have a PR with single focused green commit for review to simplify things for all the participants.
You may have a look how it works in [Eclipse Platform](https://github.com/eclipse-platform/eclipse.platform/pull/)

merks:
EGit doesn't support squash but you can use reset soft  to the commit before the 3 commits of this PR, and then create a new commit, force pushing it to the same fork branch.

---

The change to avoid using org.apache.commons.codec makes sense.  None of the httpcomponents changes make sense and are only done as sort of a workaround for PDE launch problems in the IDE.  If the codec avoidance is complete, and is not dragged in by another target platform dependency, then I expect the problem resolution problem in the target platform which is the cause for the PDE launch problem would also go away.  I'm just testing if that's true now...

gnl42:
The codec issue only showed up at the very end when the site was being built (hence removing it from category.xml)

merks:
The point is that the old httpclient requires the old codec

image

**Nothing** you do will eliminate that problem, other than to stop using httpclient .  Even removing your own use of codec will not solve that problem, but it's a good thing to use the JDK instead of codec, so I suggest doing that as it's own PR.   But again, even after you do, you can't remove it from your target platform and should not remove it from your update site because it's really required.

---

The problem is the newer commons-codec in your TP.  It appears that jgit requires the newer commons-codec:

image

Again, there is nothing you can do to avoid that.

But your test does not need to include the newer one in the launch:

image

I really don't think you should make a bunch of changes based on trying to solve a problem for launching the test in the IDE.  You're not really solving any problems, simply changing the problem and removals form the TP or the update site make no sense are are actually creating a potential new problem.

merks:
> The codec issue only showed up at the very end when the site was being built (hence removing it from category.xml)

Note that I have no idea what you mean by "the codec" issue.   That's part of the problem with this PR is there seem to be 2-4 issues wrapped into one...

gnl42:
After I made the httpcomponents change, the github build failed with:
```
Error:  Cannot resolve dependencies of project org.eclipse.mylyn:org.eclipse.mylyn-site:eclipse-repository:4.11.0-SNAPSHOT
Error:   with context {osgi.os=macosx, osgi.ws=cocoa, org.eclipse.update.install.features=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
Error:    Software being installed: org.eclipse.mylyn-site raw:4.11.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):4.11.0-SNAPSHOT
Error:    Missing requirement: org.eclipse.mylyn-site raw:4.11.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):4.11.0-SNAPSHOT requires 'org.eclipse.equinox.p2.iu; org.apache.commons.codec 0.0.0' but it could not be found: See log for details
Error:  -> [Help 1]
```

After I removed the codec references there was a clean build.

One change led to another. In the end all were needed to get a clean PDE environment and clean build

merks:
I'm so confused.  😢 

What is "the httpcomponents" change?   It would be much easier to understand if there were a PR that spells it out.  But, as I suggested, changing a package import to a bundle requirement doesn't really change anything in a semantically significant way except maybe how PDE computes the launch bundles, perhaps forcing it to include httpclient even though the target platform state indicates it has unresolved requirements (because of package substitutions of codec packages for from 1.14 to the 1.20 version from commons-codec bundle).

---

The "remove codec references" change is also unclear to me.  I think that's this subset of 3 changed:

image

Better would be to make the above simple change and build it.  When it comes to dependencies, less is more.

But in this PR you are also removing it from the category.xml while httpcomponents.httpclient definitely requires it, so it's unclear to me why/how  that's necessary and how that doesn't actually make the p2 update site incomplete.  Maybe because you are removing httpcomponents from the target you're haven a problem, even though you are explicitly adding bundle requirements to it:

image

The why for this kind of inexplicable to me.

---

A point of confusion here is that a bunch of changes are combined along with "change setup to set system property to avoid workspace errors".

What's most confusing of all is that a build against master's head is successful with no changes:

- https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/
- https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/766/

That suggests to me that no changes are needed for a successful build.  

Am I overlooking something fundamental?

merks:
My Mylyn workspace with the latest state of main (along with only the 3 codec changes) has only this one strange API error:

image

So even that part does not appear to require any changes.

gnl42:
@merks Try running CommonHttpClientTest

gnl42:
@merks  Are you running 4.40.0-M1?

merks:
Yes, M1:

image

Yes, the tests run in the IDE as I explained previously:

image

As I've also explained previously, configuring the launcher with the necessary bundles is painful but possible.

This is manually included:

image

Here I manually included the old codec and disabled the newer one:

image

When both codec versions are included it complains and doesn't work

image

---

But now that I look carefully at SimRel, JGit, and your target platform,  I notice you are using 4.5.13 in the TP:

image

And that has different crappy dependencies compared to the 4.5.14 version in the latest Orbit, JGit, and SimRel.

Moving it in the TP (or removing it and letting it transitively come from EGit/JGit) like this:

image

picks up the newer version:

image

That one does not depend on codec at all.  Then the older codec is gone from the TP and is no longer needed in the site.

None of the package -> bundle requirement changes are needed, just a proper target platform update to the 4.5.14 version of the 2 httpcomponents bundles.

I created this PR with that approach:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/1049

gnl42:
Fixed by a4fa8c6 and #1050

PR #1049, opened 2026-04-19
Stop using org.apache.commons.codec

merks:
- Ensure the latest version of org.apache.httpcomponents.httpclient is in the target platform.
- Replace org.apache.commons.codec with org.apache.commons.commons-codec in the site; it's required by org.apache.commons.httpclient.


merks:
@gnl42  @ruspl-afed 

Note that this is one of the reasons why I really dislike manually maintained target files.  After a while, you don't actually know what is coming from where and don't notice when you drag in old versions of things for reasons that are no longer applicable.

merks:
The target platform state has no unresolved plugins now.

merks:
@gnl42 

I'm headed to Brussels soon.  Probably this will address your IDE problem.  If so please merge.  I'll check mail from my phone...

ruspl-afed:
> I'm headed to Brussels soon.

Unfortunately, I haven't had this opportunity for several years, but I wish you a great conference!

ruspl-afed:
> Note that this is one of the reasons why I really dislike manually maintained target files. After a while, you don't actually know what is coming from where and don't notice when you drag in old versions of things for reasons that are no longer applicable.

And what is an alternative, that solves all the described issues @merks ?

merks:
For many projects (like BIRT) I use targlets and generate the TP.  I’ll try to find some time to experiment with that.  Several things looked likely to be outdated. E.g., commons.io versus common.commons-io.  There’s no panacea solution though; it’s a complex problem.

ruspl-afed:
> For many projects (like BIRT) I use targlets and generate the TP

TBH I do not see notable benefits in [BIRT](https://github.com/eclipse-birt/birt/blob/master/build/org.eclipse.birt.target/org.eclipse.birt.target.target) approach: a lot of repositories and units in one location and no idea how to figure out where this or that IU comes from.

merks:
Even in the current TP, it not clear where many things come from because they come from planner mode pulling in the transitive dependency from one of the specific locations  The generator does support separate locations; that actually looks nicer in a editor..  Anyway, it’s probably pointless to invest time if manual is preferred.  But who will revisit the outdated orbit dependencies?  Probably no one until the next totality confusing problem, and then it will be me. 😮

gnl42:
Hi @merks looks good for the plugin issue, still need -Dpde.addTransitiveDependenciesWithForbiddenAccess=false though to make Eclipse happy.

Should I extract my Mylyn.setup change and create a new PR just for it?

merks:
> Should I extract my Mylyn.setup change and create a new PR just for it?

Yes, please.   I still hope @HannesWell or @laeubi will have time to look at the underling problem that the new behavior causes in this case such that we can revert the change later.

PR #1052, opened 2026-04-20
Disable 'The package xyz is accessible from more than one module' error

merks:
- Add org.eclipse.jdt.core.compiler.ignoreUnnamedModuleForSplitPackage=enabled to projects with javax.xml errors to suppress those errors.
- Remove -Dpde.addTransitiveDependenciesWithForbiddenAccess=false because it's no longer needed.

https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1044


merks:
@gnl42 @ruspl-afed 

I think this solves the remaining Mylyn problems that surfaced for PDE M1.

merks:
I'll interpret no comment as agreement.

ruspl-afed:
> I'll interpret no comment as agreement.

Thank you for your effort @merks , I was waiting for comments from @gnl42

PR #1060, opened 2026-04-27
Added support for admonition paragraphs

azoitl:
For admonition paragraphs the correct preview is created.

### DIFF
--- a/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/AdmonitionParagraphBlock.java
+++ b/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/AdmonitionParagraphBlock.java
@@ -1,0 +1,110 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Johannes Kepler University Linz
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alois Zoitl - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.mylyn.wikitext.asciidoc.internal.block;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.mylyn.wikitext.parser.Attributes;
+import org.eclipse.mylyn.wikitext.parser.DocumentBuilder.BlockType;
+
+/**
+ * AsciiDoc Admonition paragraph block
+ */
+public class AdmonitionParagraphBlock extends ParagraphBlock {
+
+	private static final Pattern PATTERN = Pattern.compile(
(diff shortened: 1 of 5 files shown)


mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/ParagraphBlock.java
```diff
@@ -26,7 +26,7 @@
  */
 public class ParagraphBlock extends Block {
 
-	private int blockLineCount = 0;
+	protected int blockLineCount = 0;
```
ruspl-afed: `private` mutable field is not perfect, but `protected` one introduces a lot of future adventures. I would create a protected final method instead.

mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/AdmonitionParagraphBlock.java
```diff
@@ -0,0 +1,106 @@
...
+	@Override
+	protected int processLineContent(String line, int offset) {
+		if (blockLineCount == 0) {
+			return handleFirstLine(line, offset);
+		}
+
+		return super.processLineContent(line, offset);
+	}
+
+	private int handleFirstLine(String line, int offset) {
```
ruspl-afed, line 50: they say that if empty lines are required to separate the code for better understanding, it means that these parts should be separate methods.

mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/ParagraphBlock.java
azoitl: I was also not happy with that change. Would a getter and setter be fine?

ruspl-afed: I'm thinking about something like
```Java
protected final void incrementBlockLineCount() {
   blockLineCount++;
}
```

azoitl: ok. Then I would also add a `isFirstLine`

azoitl: As I most probably need the line count for adminition blocks later I added a `getBlockLineCount()` the rest I did accordingly.

mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/src/main/java/org/eclipse/mylyn/wikitext/asciidoc/internal/block/AdmonitionParagraphBlock.java
azoitl: did that.

ruspl-afed (approved):
LGTM

Issue #1075, opened 2026-04-29
Prepare 4.11.0 M2 contribution


ruspl-afed:
align target with for SimRel 2026-06 M2

PR #1076, opened 2026-04-29
[#1075] Prepare 4.11.0 M2 contribution

ruspl-afed:
* ECF
* EMF
* Platform
* CDT

### DIFF
--- a/org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
+++ b/org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
@@ -20,7 +20,7 @@
 			<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/eclipse/updates/4.40-I-builds/I20260402-1800/"/>
+			<repository location="https://download.eclipse.org/eclipse/updates/4.40-I-builds/I20260423-1800/"/>
 			<unit id="org.eclipse.equinox.sdk.feature.group" version="0.0.0"/>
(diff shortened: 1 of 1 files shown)


azoitl (commented):
@ruspl-afed I understand way to little about the whole build process and which locations are to be used when. But in Eclipse GEF Classic we use `https://download.eclipse.org/staging/2026-06` (which may be totally wrong). With that we only have to change that once per release cycle. Reducing a lot of maintenance effort. Maybe you can use that to for at least platform, emf, and ecf.

merks:
The only concern with staging repositories is that they are overwritten in place so it has a small potential to break a build.  In any case, both EMF and ECF have highly stable APIs...

ruspl-afed:
> @ruspl-afed I understand way to little about the whole build process and which locations are to be used when. But in Eclipse GEF Classic we use `https://download.eclipse.org/staging/2026-06` (which may be totally wrong). With that we only have to change that once per release cycle. Reducing a lot of maintenance effort. Maybe you can use that to for at least platform, emf, and ecf.

I used to do so, until it appeared that I need to understand and control better where I get particular IU from.

In other words, as a SimRel participant I need to rely on concrete UI versions that will form upcoming SimRel staging, and not on its previous version (that I will get from existing https://download.eclipse.org/staging/2026-06)
The only exception is Orbit that publishes its milestone before everything else.

However, for SimRel consumer (like 4diac) it is fine to use https://download.eclipse.org/staging/2026-06

ruspl-afed:
> so it has a small potential to break a build

and, since I have already had experience realizing this potential, I prefer to use original ingredients.

azoitl:
@merks @ruspl-afed thx for the feedback. I didn't want to be the "smart aleck" here. I was mostly curios and try to understand. AFAIK in GEF Classic it sofar worked out nicely for us. But with your feedback I fear we may need to reconsider.

merks:
It's always reasonable to questions about why things are the way they are!

PR #1089, opened 2026-05-02
Need mojibake with Bugzilla 5.x for attachment name #860

### LINKED ISSUE #860: BugzillaAttachmentHandlerTest::testAttachmentWithUnicode fails with Bugzilla 5.2
Unit test fails against Bugzilla 5.2:
```
Test failed: testAttachmentWithUnicode()
org.opentest4j.AssertionFailedError: expected: <testçñ¥€£½¼βθ台北ゖ゗ｳ1756190655668.txt> but was: <testÃ§Ã±Â¥â¬Â£Â½Â¼Î²Î¸å°åããï½³1756190655668.txt>
```

Fine with Bugzilla 4.4.
Ubuntu 25

gnl42:
Threw Copilot at the problem:
```
Bugzilla 5.x uses a Perl MIME multipart parser (from CGI.pm) to read uploaded form data. This parser does not support RFC 5987 filename* in multipart body Content-Disposition headers — it only understands the plain filename parameter. So Bugzilla 5.x still read the raw UTF-8 bytes as Latin-1 and stored the mojibake.

Root cause (confirmed by the failure output)

expected: testçñ¥€£½¼βθ台北ゖァｳ゗...txt
was: testÃ§Ã±Â¥â¬Â£Â½Â¼Î²Î¸å°åãã¡ï½³ã...txt

This is byte-for-byte mojibake: every non-ASCII character in the original has its UTF-8 bytes present in the returned string interpreted as individual Latin-1 characters. E.g., ç (U+00E7) → UTF-8 bytes 0xC3 0xA7 → Latin-1 chars Ã§ (U+00C3 U+00A7).

The fix: SaxMultiBugReportContentHandler.java

When parsing the <filename> element, the new recoverMojibakeFilename() method:

	1. Takes the returned string's bytes as ISO-8859-1
	2. Tries to decode those bytes as strict UTF-8 (failing on any invalid byte sequence)
	3. If the decode succeeds and the result differs from the input → it was mojibake → use the recovered string
	4. If the decode fails (invalid UTF-8) → filename was not mojibake → use original

This is safe because legitimate Latin-1 filenames almost never have byte patterns that form valid UTF-8 multi-byte sequences. For example:

	•  "Ñoño.txt" → ISO-8859-1 bytes [D1 6F F1 6F ...] → 0xD1 needs a UTF-8 continuation but 0x6F (o) is not one → invalid UTF-8 → no recovery applied ✓
	•  "testÃ§.txt" (mojibake) → ISO-8859-1 bytes [... C3 A7 ...] → valid UTF-8 sequence → recovered to testç.txt ✓
	•  Pure ASCII filenames → recovery produces identical string → original returned unchanged 

```

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/860

### DIFF
--- a/mylyn.tasks/connectors/bugzilla/org.eclipse.mylyn.bugzilla.core/src/org/eclipse/mylyn/internal/bugzilla/core/SaxMultiBugReportContentHandler.java
+++ b/mylyn.tasks/connectors/bugzilla/org.eclipse.mylyn.bugzilla.core/src/org/eclipse/mylyn/internal/bugzilla/core/SaxMultiBugReportContentHandler.java
@@ -14,6 +14,11 @@
 
 package org.eclipse.mylyn.internal.bugzilla.core;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
+import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -392,7 +397,7 @@ public class SaxMultiBugReportContentHandler extends DefaultHandler {
 				break;
 			case FILENAME:
 				if (attachment != null) {
-					attachment.setFileName(parsedText);
+					attachment.setFileName(recoverMojibakeFilename(parsedText));
 				}
 				break;
 			case CTYPE:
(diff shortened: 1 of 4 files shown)


merks (approved):
Seems okay though I suppose it would be good to have an actual test.  I also wonder about what might be an example of a rare case where the bytes could be misinterpreted...

gnl42:
I'll add some tests

merks (approved):
Good to have tests. Thanks!

ruspl-afed (approved):
LGTM

Issue #567, opened 2024-07-04
Update transitive com.google.gerrit dependencies


merks:
While investing https://github.com/eclipse-simrel/simrel.build/issues/438 I noticed that these very old dependencies are being pulled into SimRel by Mylyn:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/cd99e897-e287-429e-86d8-a69ebcdce8a6)

Specifically via these dependencies:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/e58a3aba-f867-465e-85c0-19bfc808ef24)

Searching the source, I see these bundles import the packages:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/a728db11-1dea-4265-9b79-4e565179e630)

I can find what might be updates/replacements for the packages here:

- https://repo1.maven.org/maven2/com/google/gerrit/gerrit-extension-api/
- https://repo1.maven.org/maven2/com/google/gerrit/gerrit-plugin-api/
- https://repo1.maven.org/maven2/com/google/gwt/gwt-user/
- https://repo1.maven.org/maven2/com/google/gerrit/gwtjsonrpc/
- https://repo1.maven.org/maven2/com/google/gerrit/gwtorm/

But I don't know that anything from the above is compatible and some of their jars look like dumping grounds that include dependencies, like this one that looks like a dependency closure:

![image](https://github.com/eclipse-mylyn/org.eclipse.mylyn/assets/208716/177a8c7b-f330-4d96-b2a8-d49fec1f0834)

I'm concerned of the risk of using 13 year old libraries that work with passwords for authentication for which we have essentially no ability to update to newer versions to fix any security problems that may exist or may arise.

I'm fine to help add newer bundles to Orbit, but I'm at a bit of a loss which things are actually need and would provide suitable updated alternatives.

The 3.9.5 version of gerrit-extension-api provides quite a bit and doesn't have so many dependencies, though I've made then all optional here in order to build the jar:
```
Export-Package                          com.google.auto.value.extension.memoized;version="3.10.0"
                                        com.google.auto.value.extension.serializable;version="3.10.0"
                                        com.google.auto.value.extension.toprettystring;version="3.10.0"
                                        com.google.auto.value;version="3.10.0"
                                        com.google.common.annotations;version="3.10.0"
                                        com.google.common.base.internal;version="3.10.0"
[...]

Issue #74, opened 2023-04-01
Drop support for Gerrit Connector


BeckerFrank:
Here the reasons.
- Eclipse is moving from git / Gerrit to GitHub / Gitlab
- the actual active Gerrit version is 3.5 or greater (see [Gerrit Supported Versions](https://www.gerritcodereview.com/support.html))
- the mylyn.org instances with version 2.11.10 or 2.13.8 do not work anymore (see Issue #26 we have failing Gerrit tests)

As a first step we remove the Gerrit Connector from the next Mylyn version (3.26).

BeckerFrank:
The next step, deleting the source, is planned for after 3.26.

wimjongman:
Can we/Should we isolate the connector in a separate repo?

ruspl-afed:
Perhaps we should not. I consider the drop of Gerrit support as a decision forced by limited resources.
However mylyn.gerrit subfolder could be good.

PR #1100, opened 2026-05-07
Replace IOUtils, part1 #1098

### LINKED ISSUE #1098: Replace IOUtils
IOUtils should be replaced by native java

gnl42:
First half of replacing IOUtils with native java

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1098


mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
```diff
@@ -15,7 +15,8 @@ Export-Package: org.eclipse.mylyn.commons.core,
  org.eclipse.mylyn.internal.commons.core;x-internal:=true,
  org.eclipse.mylyn.internal.commons.core.operations;x-internal:=true
 Import-Package: org.apache.commons.collections4.multimap;version="4.4.0",
- org.apache.commons.text;version="[1.12.0,2.0.0)"
+ org.apache.commons.text;version="[1.12.0,2.0.0)",
+ org.eclipse.core.resources
```
ruspl-afed: this is relatively heavy dependency, it is better to avoid it.
please consider splitting your utility and move `IFile`-dependent part somewhere else

gnl42: Can you expand on that please?

mylyn.reviews/org.eclipse.mylyn.gerrit.tests/src/org/eclipse/mylyn/internal/gerrit/core/remote/GerritDataLocatorTest.java
```diff
@@ -52,10 +58,7 @@ void skipIfExcluded() {
 	@BeforeEach
 	void setUp() throws Exception {
 		File rootDir = new File(locator.getSystemDataPath().toOSString());
-		if (rootDir.exists()) {
-			FileUtils.forceDelete(rootDir);
-		}
-		FileUtils.forceMkdir(rootDir);
```
merks, line 58: No more mkdir?

merks (commented):
It's a lot to review, and the formatting adds to the noise.  

In principle it's good, but bad things might be hiding in the details.

merks:
It looks like the build failure was an infrastructure problem.  I tried restarting it just now.

mylyn.commons/org.eclipse.mylyn.commons.core/META-INF/MANIFEST.MF
merks: Because it makes this bundle that does not depend on a workspace depend on a workspace simply for one utility method. Do we know that no one uses this bundle in a context without a workspace?  (In EMF we are very careful about what does and doesn't work with and without a workspace.)

ruspl-afed: Currently `org.eclipse.mylyn.commons.core` is relatively lightweight bundle, while adding `org.eclipse.core.resources` implies the whole Eclipse Resources subsystem to be initiated, including workspace creation and so on. I don't think that we need to add such a heavy requirement for all the downstream Mylyn bundles only to expose 1 utility method for `IFile`. There must be a better place for such a method if it is really needed.

gnl42: It's only needed in four places. I'll inline the code, and remove the method

mylyn.reviews/org.eclipse.mylyn.gerrit.tests/src/org/eclipse/mylyn/internal/gerrit/core/remote/GerritDataLocatorTest.java
gnl42: oops, fixed. 

What is going on with the gerrit connector? It was partially removed from category.xml with #74, not sure when it was added back.

Never did hear anything on #511

ruspl-afed (approved):
LGTM

mylyn.reviews/org.eclipse.mylyn.gerrit.tests/src/org/eclipse/mylyn/internal/gerrit/core/remote/GerritDataLocatorTest.java
ruspl-afed: > Never did hear anything on #511

fixed

ruspl-afed: > gerrit connector

For me, the key question is what to connect to with it and whether it will be possible for me with all the connection restrictions that we are currently facing in different regions.

gnl42: There is also the discussion in #567 about maybe splitting out the gerrit, and some other connectors, into it's own repository

ruspl-afed: > oops, fixed.

@merks do you have more comments regarding this?

Issue #1106, opened 2026-05-22
Little progress update shown for Github query


gnl42:
When retrieving issues from Github, the progress bar doesn't show any progress as the issues are retrieved.
It only updates for the "statuses".

The query also can't be cancelled.

PR #1107, opened 2026-05-22
Little progress update shown for Github query #1106

gnl42:
Use subMonitor.split(1) to:

1. Show progress as issues are processed from Github
2. Make it possible to cancel the query

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1106


merks (approved):
Looks okay but the best judge is setting breakpoints to see how it actually looks.

ruspl-afed (approved):
LGTM

PR #1112, opened 2026-05-23
Unstable Jenkins CI server #1111

### LINKED ISSUE #1111: Unstable Jenkins CI server
The Jenkins CI server goes down every once in a while, causing the github builds to fail.

Skip the jenkins CI unit tests if the server is unavailable

gnl42:
Skip CI unit test if no Jenkins fixture are found.

May not be the best solution, but it stops PR builds from failing due to infrastructure problems.

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1111

### DIFF
--- a/mylyn.builds/org.eclipse.mylyn.jenkins.tests/src/org/eclipse/mylyn/jenkins/tests/AbstractFixtureTest.java
+++ b/mylyn.builds/org.eclipse.mylyn.jenkins.tests/src/org/eclipse/mylyn/jenkins/tests/AbstractFixtureTest.java
@@ -11,7 +11,7 @@
 
 package org.eclipse.mylyn.jenkins.tests;
 
-import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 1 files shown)


merks (commented):
Seems okay, but on the downside the server could start failing forever and then who would notice it need attention?

gnl42:
Yeah,  but it has been down for a couple of weeks now and nothing seems to have been done to fix it.

I was wondering if there was some way to trigger an alert of some sort during the build that could be reported?

ruspl-afed:
> it has been down for a couple of weeks now and nothing seems to have been done to fix it.

true. 
I'm trying to update SimRel regularly, but exploring the root cause more deeply is beyond my current capabilities.
And, the connection in my region does not always provide effective access to the necessary network resources.

ruspl-afed (approved):
LGTM
This is the only way for now to unblock Mylyn pipeline

PR #1119, opened 2026-05-31
Replace IOUtils, part 2 #1098

gnl42:
Replace some more IOUtils usage

Should commons.core.FileUtil be referenced like this???

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1098


ruspl-afed:
> Should commons.core.FileUtil be referenced like this???

TBH, I would inline these usages and leave only standard java io/nio calls, even if it's a little more verbose

ruspl-afed:
I was not expecting GitHub to create a merge commit for this change, so I plan to force-push the conflict resolution changes to this branch.

merks:
You have to be careful when you update the branch via the webpage that it defaults to merge and the rebase choice while available not the default.

ruspl-afed:
> You have to be careful when you update the branch via the webpage that it defaults to merge and the rebase choice while available not the default.

It was a bit different UI, available after solving conflicts.

ruspl-afed (approved):
I would merge it as it is, to avoid further merge conflicts.

PR #1124, opened 2026-06-03
Update to avoid errors about forbiddenReference set to warning

### LINKED ISSUE #1123: Eclipse 4.40 RC1 issues:
Seeing these type of errors from Eclipse after the 4.40 RC1 upgrade:

- API analysis aborted for 'org.eclipse.mylyn.bugzilla.ide' since its build path is incomplete
- OSGi access rules not enforced by classpath access rules, which can lead to classloading errors at runtime.

merks:
- Add package imports to eliminate forbiddenReference errors

Fixes https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1123


ruspl-afed:
A note for upcoming 2026-06:
from the nature of suggested changes it looks more like PDE problem, that has no impact on runtime functionality (except 2 changes in tests MANIFEST.MF, that are not a part of SimRel contribution).
So, I think we don't need a re-spin of Mylyn because of this issue.

merks:
Definitely no need to respin!   It's mostly an new error generated by PDE "for your safety and comfort". 

So these are just changes you need for the next cycle to get the workspace in working order.

merks:
@gnl42 @ruspl-afed 

I'll leave this with you to merge if/when you are comfortable with that.

gnl42:
Seems to work

PR #1149, opened 2026-06-15
Make o.e.mylyn.wikitext.parser.markup API

azoitl:
By turning this package API clients of Wikitext can extend existing markup languages with own language elements. An example can be links to model elements for which the markdown file is a documentation.


ruspl-afed:
The change is indeed laconic, but not so trivial, since it promotes to an API a number of ancient types that may require to open even more related types to complete the scenario. 
Personally I tend to keep them hidden and open a bit more modern and solid API, however your live interest to the subject may change my opinion @azoitl 
What is missed here in the PR is a reference to a problem that we are going to solve with this PR, because here you are talking about "own language elements", while in the discussion you are asking about "how to create an instance of `HtmlLanguage`"
It would be great to ensure that this change is required and sufficient to solve your integration problem and not just a first step to turn too much code around into an API.
In other words, can we have [a reference to] a desired code snippet?

azoitl:
@ruspl-afed I understand your confusion and my message was way to short. I think I should first point out my use-cases:

1. I would like to embedded an wikitext markdown editor (source and preview) into an editor tab of a multipage editor.
2. I would like add custom parser rules with custom preview to an existing markdown language (in our case Asciidoc)
3. I would like to have a HTML preview window for markdown files for a certain subsets of my editors (i.e., read only files)

For 1 and 3 I would at least need to be able to hold a variable of type `MarkupLanguage` as the language I would like to parse has this as baseclass as well as the `HtmlLanguage` for the HTML preview.

For 2 I need to create an own `MarkupLanguageConfiguration` with its associated classes. Mainly `Block` and `PatternBasedElement`/`PatternBasedElementProcessor`. 

I created a first prototype for testing using a lot of internal API. The two main classes are the 

- Editor: https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.fbtypeeditor.asciidoc/src/org/eclipse/fordiac/ide/fbtypeeditor/asciidoc/editors/AsciiDocDocumentationEditor.java
- Viewer: https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.fbtypeeditor.asciidoc/src/org/eclipse/fordiac/ide/fbtypeeditor/asciidoc/editors/AsciiDocPreviewTypeEditorPage.java

I think in the second file you see it a bit better. It is the code starting from line 153.

Most probably I missed a lot and I'm happy for any suggestions to make this better.

ruspl-afed (approved):
After reviewing the code around it is clear for me that the most straightforward way for today is to let this former API return, otherwise we will loose another usage that has a potential to be a driver for further Mylyn Docs evolution.
May be a few of related packages will follow.

PR #1154, opened 2026-06-16
Use @SetSystemProperty in unit tests #1151

### LINKED ISSUE #1151: Use @SetSystemProperty
Replace System.setProperty in unit tests with @SetSystemProperty

gnl42:
Use @SetSystemProperty annotation instead of System.setProperty in unit tests

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1151


merks (approved):
Looks cleaner.

ruspl-afed (approved):
This may lead us to a philosophical discussion about semantic expressiveness and so on, that I would love to have with all of you in person one day.
OK, setting system property "outside" is much better than doing it "inside".

Issue #1156, opened 2026-06-17
JAXBContext ClassNotFoundException JAXBContextFactory


gnl42:
Maven runs after #1152  of the jenkins tests produces some 300k lines of the format:
```
Running JenkinsClientTest.testAbortBuild@JENKINS 2.506
java.lang.ClassNotFoundException: org.glassfish.jaxb.runtime.v2.JAXBContextFactory cannot be found by org.glassfish.jaxb.runtime.source_4.0.9
	at org.eclipse.osgi.internal.loader.BundleLoader.generateException(BundleLoader.java:570)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass0(BundleLoader.java:564)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:439)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:195)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.loadClass(EquinoxBundle.java:663)
	at org.glassfish.hk2.osgiresourcelocator.ServiceLoaderImpl.loadClassSecured(ServiceLoaderImpl.java:151)
	at org.glassfish.hk2.osgiresourcelocator.ServiceLoaderImpl.lookupProviderClasses1(ServiceLoaderImpl.java:123)
	at org.glassfish.hk2.osgiresourcelocator.ServiceLoader.lookupProviderClasses(ServiceLoader.java:109)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at jakarta.xml.bind.ServiceLoaderUtil.lookupUsingOSGiServiceLoader(ServiceLoaderUtil.java:60)
	at jakarta.xml.bind.ContextFinder.find(ContextFinder.java:371)
	at jakarta.xml.bind.JAXBContext.newInstance(JAXBContext.java:605)
	at jakarta.xml.bind.JAXBContext.newInstance(JAXBContext.java:546)
	at org.eclipse.mylyn.internal.jenkins.core.client.RestfulJenkinsClient.unmarshal(RestfulJenkinsClient.java:452)
```
ad nauseam,

[...]

PR #1157, opened 2026-06-20
Add bundle require for jakarta.xml.bind-api to o.e.mylyn.jenkins.core

merks:
https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1156


gnl42:
The problem  is still there
```
java.lang.ClassNotFoundException: org.glassfish.jaxb.runtime.v2.JAXBContextFactory cannot be found by org.glassfish.jaxb.runtime.source_4.0.9
[2](https://github.com/eclipse-mylyn/org.eclipse.mylyn/actions/runs/27875837087/job/82495058586#step:5:25858)
```

merks:
Yes, none of my efforts helped so far.   😢 

I wonder if it's deploying the source bundles in the launch?  A launch from the IDE doesn't have this problem.

I wanted to analyze all the orbit bundles to see which one(s) have source bundles with questionable exports.  Because I'm not 100% sure which thing is causing which problem, or even if for sure it's the same problem...

merks:
I didn't realize is the source bundles with service registration that cause problem, and I didn't realize there are so many such things!!!

PR #1159, opened 2026-06-22
Update from org.apache.commons.logging 1.2 to org.apache.commons.commons-logging 1.4

merks:


merks:
@ruspl-afed  @gnl42 

This appears to have been successful.  Any concerns about merging this?

ruspl-afed (approved):
LGTM
I saw that we in fact grab 1.2.0 from some old repo, when I was trying to update Gerrit, but I was not sure.
Thank you @merks

merks:
BTW, given that the target platform uses planner mode and given that only the newer logging bundle is present in the target platform, I was quite hopeful that this was the correct thing to do.  :-)

image

gnl42:
Thanks @merks

PR #1163, opened 2026-06-23
[#1162] Resolve `@since` warnings for `org.eclipse.mylyn.jenkins.core`

### LINKED ISSUE #1162: Resolve `@since` warnings for `org.eclipse.mylyn.jenkins.core`
Extract from https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/1158#issuecomment-4776695982 (by @merks ) 

And it's hard to work with the offending project because it has so many complaints about missing `@since`. Don't others see that?

Image

ruspl-afed:
* internal packages should not be a part of API

Fixes #1162

### DIFF
--- a/mylyn.builds/org.eclipse.mylyn.jenkins.core/META-INF/MANIFEST.MF
+++ b/mylyn.builds/org.eclipse.mylyn.jenkins.core/META-INF/MANIFEST.MF
@@ -14,9 +14,9 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Localization: plugin
-Export-Package: org.eclipse.mylyn.internal.hudson.model,
- org.eclipse.mylyn.internal.jenkins.core,
- org.eclipse.mylyn.internal.jenkins.core.client;x-friends:="org.eclipse.mylyn.jenkins.ui",
+Export-Package: org.eclipse.mylyn.internal.hudson.model;x-friends:="org.eclipse.mylyn.jenkins.tests",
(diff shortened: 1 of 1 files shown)


ruspl-afed:
We have a process issue here: 
formally this is a breaking API change, that requires promotion to 5.0
but in fact it looks more like a fix for "API by mistake", because packages have "internal" in its name.

mylyn.builds/org.eclipse.mylyn.jenkins.core/META-INF/MANIFEST.MF
```diff
@@ -14,8 +14,8 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Localization: plugin
-Export-Package: org.eclipse.mylyn.internal.hudson.model,
- org.eclipse.mylyn.internal.jenkins.core,
+Export-Package: org.eclipse.mylyn.internal.hudson.model;x-internal:=true,
```
merks: Is it used in any tests that should be friends? Friend directives automatically imply internal too.

ruspl-afed: Fixed

merks (approved):
Yes. That looks better.

gnl42 (approved):
Looks good.

My question is, why/how did the internal classes become API? Especially since they weren't touched by #1153

ruspl-afed:
> My question is, why/how did the internal classes become API? Especially since they weren't touched by #1153

`x-friends` directive was removed from package => no more constraints => API

I asked @akurtakov for review to confirm that this side effect approval was not intentional

akurtakov (approved):
From my side I missed that part of the change while doing the review. Sorry about that!

ruspl-afed:
> From my side I missed that part of the change while doing the review. Sorry about that!

No problem!

PR #1164, opened 2026-06-23
Simplify/improve the target platform and the update site

merks:
- Convert org.eclipse.mylyn.commons.notifications.feed to jakarta.xml.bind like org.eclipse.mylyn.jenkins.core.
- Generate a new org.eclipse.mylyn.target.target with fewer old Orbit update sites.
- Modify the category.xml to include updated Orbit dependencies.

### DIFF
--- a/org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
+++ b/org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
@@ -1,108 +1,108 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<!--
- *******************************************************************************
- * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * https://www.eclipse.org/legal/epl-2.0/.
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *   See git history
- *******************************************************************************
---><target name="org.eclipse.mylyn.target">
-	<locations>
-		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/cbi/updates/license/"/>
-			<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/eclipse/updates/4.41-I-builds/"/>
-			<unit id="org.eclipse.equinox.sdk.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.equinox.p2.sdk.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.sdk.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/egit/github/updates/"/>
-			<unit id="org.eclipse.egit.github.core" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/egit/updates-7.7/"/>
-			<unit id="org.eclipse.egit.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.jgit.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.jgit.lfs.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/modeling/emf/emf/builds/release/2.46.0/"/>
-			<unit id="org.eclipse.emf.sdk.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/rt/ecf/3.16.8/site.p2/3.16.8.v20260415-2025/"/>
-			<unit id="org.eclipse.ecf.remoteservice.sdk.feature.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/cdt/builds/12.5/cdt-12.5.0-rc2/"/>
-			<unit id="org.eclipse.cdt.platform.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.cdt.sdk.feature.group" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20180206163158/repository/"/>
-			<unit id="com.google.gerrit.common" version="0.0.0"/>
-			<unit id="com.google.gerrit.prettify" version="0.0.0"/>
-			<unit id="com.google.gerrit.reviewdb" version="0.0.0"/>
-			<unit id="com.thoughtworks.qdox" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20221123021534/repository/"/>
-			<unit id="jakarta.servlet" version="0.0.0"/>
-			<unit id="org.apache.xmlrpc.client" version="0.0.0"/>
-			<unit id="org.apache.xmlrpc.common" version="0.0.0"/>
-			<unit id="org.apache.xmlrpc.server" version="0.0.0"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20230531010532/repository/"/>
-			<unit id="com.sun.xml.bind" version="2.3.3.v20221203-1659"/>
-			<unit id="jakarta.xml.bind" version="2.3.3.v20221203-1659"/>
-			<unit id="javax.activation" version="1.2.2.v20221203-1659"/>
-			<unit id="javax.activation" version="2.0.0.v20221203-1659"/>
-		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2026-09/"/>
-			<unit id="junit-jupiter-api" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-jupiter-engine" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-jupiter-migrationsupport" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-jupiter-params" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-platform-commons" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-platform-engine" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-platform-launcher" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-platform-suite-api" version="[6.0.0,7.0.0)"/>
-			<unit id="junit-platform-suite-engine" version="[6.0.0,7.0.0)"/>
-			<unit id="org.apiguardian.api" version="0.0.0"/>
-			<unit id="org.glassfish.hk2.osgi-resource-locator" version="0.0.0"/>
-			<unit id="com.sun.xml.bind.jaxb-core" version="0.0.0"/>
-			<unit id="org.glassfish.jaxb.runtime" version="0.0.0"/>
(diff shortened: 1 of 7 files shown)


merks:
I used targlets to ensure the target platform is smaller and has only the things actually required by the actual projects (though I always include the Eclipse SDK so that one can do a self-hosted launch:

image

I used negative requirements to exclude outdated Orbit dependencies from the target platform that are otherwise pulled in because of package requirements are happy to resolve to those. There is only one old Orbit repository purely for the Gerrit dependencies that I did not added to the new Orbit because they are ancient and appear unmaintained.

As a side-effect of resolving the targlet, a *.target is generated.

image

The generator minimizes the list of bundles and features based on the fact that planner mode is used. and in this case generates separate locations to more closely match the previous approach.

I ran the build locally to ensure that tests pass.

Then I tried installing into a Eclipse SDK IDE with only the 4.41-I-build repository available in addition to the Mylyn repository:

image

And enhanced the content of the category.xml until the install was successful:

image

I excluded the Git (EGit) and CDT features because those require more update sites; so it's still possible something is missing, but not so likely...

merks:
Looks like the buidl gets to the site building and doesn't find `org.apache.commons.commons-logging`:

> Missing requirement: org.eclipse.mylyn-site raw:4.12.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):4.12.0-SNAPSHOT requires 'org.eclipse.equinox.p2.iu; org.apache.commons.commons-logging 0.0.0' but it could not be found: See log for details

Not sure they my local build didn't fail that way.  

Strange where things can come from!

image

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
ruspl-afed, line 1: Ordered a bit different, we (ok, I) may need a short training how to repeat target/baseline generation locally to support this form in the future

merks: Don't worry.  This is not a hit-and-run exercise.  The generator does specify to sort the location:

image

But maybe you mean the units come before the repository?  We could specify yet another option for that too... Nothing is cast in stone here.  We've slowly added more functionality, as desired..  E.g., you can specify to generate version ranges for some IUs, or exact versions for some IUs...

One of the key points for me is that we specify a source locator, which effectively induces a p2 repository (using the p2 publisher) for the sources, along with a bunch of p2 repositories; the targlet support uses the p2 director to compute a plan from those source IUs and their requirements (where the * requirements specifies to require everything from the source location, though we can filter those with expressive predicates. ) So one resolves all the actual requirements of the actual Mylyn bundles and features to what's available in a set of repositories.  Because everything is expressed as requirements, you can express a negative requirement to exclude something.  If resolution then fails, you can determine why that thing is required and by what.  I used that to reduce the things being resolved from the older Orbit repository until it was just the gerrit bundles.

Anyway, it's just a tool.  It's really hard to author a good target platform, especially for a complex problem with so many upstream dependencies!  So better to use a tool...

merks:
Misleadingly the tests all pass for me locally in the build:

image

But not really when you look in details at the log:

image

Only a bit confusing.

merks:
@ruspl-afed @gnl42 

I did some debugging of the JAXB details.  Some details are hard-coded default:

image

Goodness knows what the thinking is behind this design and how anything else ever provides an implementation:

image

In any case, it continues to look horrible to me, but the goal of this PR is to clean up and simplify the target platform, and the tests pass now  Also the manual install into SDK test still works...

gnl42:
> Misleadingly the tests all pass for me locally in the build:
> ...
> But not really when you look in details at the log:
>...
> Only a bit confusing.

-Dmaven.test.failure.ignore=true ???

merks:
Ah ha!  I didn't look closely at the args!!

image

Maybe maybe not everything will actually work for a local build on a Windows machine...

merks:
@gnl42 

Note that I with the latest force-push I converted org.eclipse.mylyn.commons.notifications.feed like what you did for org.eclipse.mylyn.jenkins.core:

- Convert org.eclipse.mylyn.commons.notifications.feed to jakarta.xml.bind like org.eclipse.mylyn.jenkins.core.

@ruspl-afed 

I see that PDE by default creates this order:

image

So I look at the generator to order things differently.  Order doesn't actually matter (isn't meanginful) and the Platform uses both orders:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b6cfdc619ff2833a7dec393dd7d546a036c9e1dc/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target#L73-L94

Go figure.

gnl42:
> Ah ha! I didn't look closely at the args!!

Yeah, drives me crazy. "Clean" build locally, jenkins build fails ~!$*&

> 
> Maybe maybe not everything will actually work for a local build on a Windows machine...

Other than one test that fails on a 5K2K monitor, I've had no problems doing local builds on a Windows machine

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
ruspl-afed: Thank you for a lot of details @merks 
May be you can share one important point: how one can generate `.target` manifest from `.setup` manifest?

ruspl-afed:
> Order doesn't actually matter (isn't meanginful) 

I know. But order does matter to minimize changes and simplify support
If tool supports it  - perfect!

> and the Platform uses both orders

I know. And I don't think that everyone should follow this practice.

org.eclipse.mylyn-target/org.eclipse.mylyn.target.target
merks: I will include the setup changes in the PR after I make some improvements to the generator and after I have properly tested which projects it imports because it's kind of a replacement (alternative to) the import task and the activation of the *.target.  And, given earlier concerns, I don't want to force this on anyone, so the setup could include both approaches, with one disabled, or even with a boolean checkbox to choose the approach. 

Anyway, more details will follow...

merks:
> > Order doesn't actually matter (isn't meanginful)
> 
> I know. But order does matter to minimize changes and simplify support If tool supports it - perfect!
> 

Yes, I'm adding an option to support switching the order or units and repositories.

> > and the Platform uses both orders
> 
> I know. And I don't think that everyone should follow this practice.

Options can make everyone happy. 😁 

---

In terms of the order of the locations, the generator can sort them.   But the order of the repositories specified in the targlet itself is relevant to the generator because the processing involves looking at each repository and determining which of the IUs in the resolved target are available in that repository, each of which is then assigned to that repository and removed from further consideration.    So we definitely want the Orbit aggregation to be processed before the old Obit repository which is there only for the Gerrit bundles.  In general it's best if the Orbit aggregation is processed before all others, but given the minimization process which computes root IUs and avoids explicitly specifying transitive dependencies, we see quite a clean simple result.

ruspl-afed:
> So we definitely want the Orbit aggregation to be processed before the old Obit repository which is there only for the Gerrit bundles.

Understood.
I have no problem with any reasonable and determined (repeatable) order of things.

merks:
This commit the repository elements are before the unit elements in the *target.

The commit also includes the new targlet tasks, but they are disabled...

Note there are docs for the target generation:

https://eclipse.dev/oomph/?file=Authoring_Guide.md#generating-pde-target-definition-files

At this point I'm not 100% sure how to bring closure to the topic.  I think manually maintaining a target file for such a complex project with so many dependencies is kind of a losing battle.

Also I'm not sure the purpose of org.eclipse.mylyn.baseline.target.  It's not used in the build is it?

ruspl-afed:
> Also I'm not sure the purpose of org.eclipse.mylyn.baseline.target. It's not used in the build is it?

Even if it is not used yet in a pipeline it is planned to be there. So, it should be processed with the same  (Oomph?) tooling.

merks:
Even the Platform doesn’t use a target file for their api baseline. Only your own content matters for api analysis so a reference to your release update site is more than sufficient. Best  to avoid maintaining yet another complex artifact that provides effectively no value.

ruspl-afed:
> so a reference to your release update site is more than sufficient.

Thank you for your advice @merks 
So I believe you suggest to drop target manifest in favor of `.setup` baseline. I'm fine with it.

azoitl:
Not sure if it helps but in Eclipse GEF classic we have our own update site with the last version and our own features in a target file. This is very low key maintenance and can be easily plugged into PDE for direct base line violation feedback during development.

merks:
In the end, the problem is simple and there are multiple ways to express the same thing.  

As targlets it's like this:

image

Specify the root Mylyn requirements, specify the repositories from which they come.   In this case `release/latest` will always be correct (for the main/master stream).  One can do effectively the identical thing in a *.target.

IN Tycho it is generally even easier where you just specify the release repository:

image
 
Tycho will know which things to look for in that baseline, so not need to list them.

The API baseline is really not a hard problem., so best not make it complicated when that brings no value.

---

I'll try to get this finalized tomorrow....

merks:
The only change since the previous commit was to the setup which doesn't affect the build, so I will ignore the Jenkins failure.

ruspl-afed:
I can see that we have two places to describe dependency (`.setup` and `.target`), that doesn't look as simplification.
However, I may miss something important.
@merks could you please explain a bit how target update process should look now?

gnl42:
@merks I just pulled main and am seeing errors:
image

Clean... makes no difference

merks:
When this task runs, the *.target is updated:

image

When you want to change the target definition, you edit this and perform the task to update the generated *.target

image

If you need help. you ask.

---

API baseline is sometimes annoying with updates/changes not taking effect, when in doubt, go to the preferences, edit the thing, refresh it, and do full build after when it asks.

image

But the build path incomplete problem looks somehow different and is not what I see:

image

merks:
FYI, I'll be out of office for a few hours with no IDE to do anything.

merks:
@gnl42 

I hope you're back on track?  For testing, I set up of a brand new IDE from scratch and it looks fine, so I don't think anything is fundamentally wrong.

image

@ruspl-afed 

One thing that is fundamentally simpler with targlets is that one doesn't actually have to list any dependencies because it reuses the dependencies already expressed in the source repository's bundle and features.  All but the sdk dependency, ironically, are exclusions to tell the p2 planner not to use those specific things in favor of alternative dependencies from the most recent Orbit repository:

image

Anyway, at any time you can edit the *.target and any time you can activate it. 

How to add new things to the target?   You can literally just add a package imports to a bundle, run the targlet task, and it will find the bundle that provides that and include it the target platform, regenerating the *.target.

What I also do sometimes in my Eclipse SDK workspace in use Navigate -> Open Setup -> Workspace and create  targlet task there (here disable):

image

To do extra fancy debug of problems, I can augment my target platform with a bunch of other things like EGit, m2e, wwd, or whatever thing I want to test against the Eclipse SDK changes...

gnl42:
@merks 
> @gnl42
> 
> I hope you're back on track? For testing, I set up of a brand new IDE from scratch and it looks fine, so I don't think anything is fundamentally wrong.

Weird. I restarted Eclipse and got a new set of issues:
image
Clean/refresh .target didn't help, so I nuked everything and re-installed.

Seems fine after a couple of restarts

> 
> Anyway, at any time you can edit the *.target and any time you can activate it.
> 
> How to add new things to the target? You can literally just add a package imports to a bundle, run the targlet task, and it will find the bundle that provides that and include it the target platform, regenerating the *.target.

Stupid question, but how can you add an import if Eclipse doesn't know anything about the new package?

merks:
> Stupid question, but how can you add an import if Eclipse doesn't know anything about the new package?

You can textually insert a package import in the source view of the  MANIFEST.MF editor.  Then perform the tasks which will add it to the target platform and to the generated target file. Or you can edit the setup, use the repository explorer to find what you need, and drag it onto the targlet, save, perform the tasks, and then use it. Once used, you can throw the set changes away because the explicit requirement is redundant.  Or you can set the .target as your active target platform and edit it similarly to add whatever you want. Once you actually use what you added and perform the setup tasks, the generator will also add it to the generated target, I.e., overwrite your changes with the normalized generated result. 

Many different ways will work. 

If you get confused, you know I generally answer questions quickly.  And I can add details to the Oomph authoring guide as needed.

gnl42:
Duh! So used to changing .target first and then searching for the package

gnl42:
@merks Maybe add a readme to the mylyn-target folder?

gnl42:
@merks Just noticed that the target editor has it's own idea as to how the .target file should be organized:
image

merks:
I always uses a nightly installer and the changing the order is a new feature.  I'll change the setup to use the nightly update site for Oomph.

merks:
Please do a pull, and perform the setup tasks, particularly the p2 task first, which will of course require a require a restart.

ruspl-afed:
@gnl42 It looks like we have no choice but to learn how to use Oomph as a primary target/baseline source. I cannot state that I'm fully comfortable with it (since I usually edit target manually to follow SimRel updates), but I believe that @merks is acting in the best interests of the project, and we will eventually achieve a positive outcome.

merks:
Personally I would find the following would easier...  

Open the setup editor, and expand to show the targlet task.  Use the repository explorer to find the bundle you want, and drag and drop it onto the task:

image

image

The explorer is advanced mode (the C, for capability icon).  You can even git it to show package capabilities via the combo that's circled.  (Sometimes bundle names and package names are quite confusingly different.)

image

With targlets you can even express "package imports" which you cannot do with a *.target file:

image

Performing the task, you can see that the generator adds it:

image

merks:
Yes, writing some documentation is on my TODO list because.

For those who wish to edit the target, that's still okay too, but you need to activate it, easily done in the target editor toolbar, and before you commit, you need to ensure your edits are also what the generator will generate, i.e., perform setup tasks.

But I will write generic docs in Oomph, because it's useful also for the Platform SDK users, and then write some docs specific to Mylyn, with references.

---

You'll note that as part of this process here we eliminated to the use of two old orbit repositories. and now it's clear that the one old orbit repository that's left is there only for the badly outdated gerrit libraries (which are just a CVE report waiting to happen).

---

Getting ahead of ourselves, what I generally do for BIRT at the end of the release cycle is "lock down the target platform".

image

I'm not sure we can really do that for Mylyn because I'm not sure all the dependencies have final links before the release.  I'm not even sure if you care to do that for Mylyn, but for BIRT folks (external uses) like to make a branch and be able to rebuild it so for that, having fixed unchanging update sites is important

ruspl-afed:
> I'm not sure all the dependencies have final links before the release.

My experience shows that we have no final links during Mylyn release: when we contribute Mylyn RC2 to SimRel RC2 only Orbit and EMF are final.

merks:
> My experience shows that we have no final links during Mylyn release: when we contribute Mylyn RC2 to SimRel RC2 only Orbit and EMF are final.

Exactly.  Whereas for BIRT I do these things **after RC2** because BIRT is not on the train.  One less releng step to worry about is simpler.  😁   Anyone who wants to build from a branch can update target URLs themselves!

Issue #770, opened 2025-08-04
ShellDragSupportTest::handleMouseMove test failure


gnl42:
Seeing this error in Windows 11 for the current main:

```
Argument(s) are different! Wanted:
shell.setLocation(
    106,
    160
);
-> at org.eclipse.swt.widgets.Shell.setLocation(Shell.java:1583)
Actual invocations have different arguments at positions [0, 1]:
shell.setLocation(
    100,
    150
);
-> at org.eclipse.mylyn.commons.ui.ShellDragSupportTest.setUp(ShellDragSupportTest.java:42)
shell.setLocation(
    Point {100, 150}
);
...
```

[ShellDragSupportTest.handleMouseMove-stackTrace.txt](https://github.com/user-attachments/files/21573702/ShellDragSupportTest.handleMouseMove-stackTrace.txt)

PR #1168, opened 2026-06-28
ShellDragSupportTest.handleMouseMove fails when Shell.zoom() !== 100

merks:
- Given the test is known to fail when the shell zoom is not 100 and given that the GHA and Jenkins buld always run with zoom 100, we could avoid local test failure avoiding to run the test when the zoom isn't 100.

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.ui.tests/src/org/eclipse/mylyn/commons/ui/ShellDragSupportTest.java
+++ b/mylyn.commons/org.eclipse.mylyn.commons.ui.tests/src/org/eclipse/mylyn/commons/ui/ShellDragSupportTest.java
@@ -14,6 +14,7 @@ package org.eclipse.mylyn.commons.ui;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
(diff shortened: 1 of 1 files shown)


merks:
@gnl42 

I ran all the Mylyn tests locally and only this one test fails and if and only if my monitor zoom != 100.  So I would like to add an assumption to avoid that failure to ensure that I can more easily verify that the full build works locally with no tests errors.

OK?

ruspl-afed (approved):
LGTM

mylyn.commons/org.eclipse.mylyn.commons.ui.tests/src/org/eclipse/mylyn/commons/ui/ShellDragSupportTest.java
```diff
@@ -27,6 +27,7 @@
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Shell;
+import org.junit.jupiter.api.Assumptions;
```
gnl42: import static?

mylyn.commons/org.eclipse.mylyn.commons.ui.tests/src/org/eclipse/mylyn/commons/ui/ShellDragSupportTest.java
```diff
@@ -82,6 +83,9 @@ public void handleMouseMove() {
 		handleEvent(SWT.MouseMove, 110, 160);
 		assertEquals(new Point(100, 150), shell.getLocation());
 
+		int zoom = shell.getZoom();
+		Assumptions.assumeTrue(zoom == 100, "This test is skipped if the monitor zoom is not 100: zoom=" + zoom); //$NON-NLS-1$
```
gnl42: Test classes have been using '@SuppressWarnings("nls")'

gnl42 (approved):
Works for me. Another version of #770

merks:
@gnl42 

Thanks for the feedback.  All good ideas and all incorporated.  Also a link to the issue in the commit.

PR #1169, opened 2026-06-28
Use minimizeLocationUnits to reduce the generated *.target content

merks:


merks:
The new option is documented here:

https://eclipse.dev/oomph/?file=Authoring_Guide.md#generating-pde-target-definition-files

This documentation was added:

https://eclipse.dev/oomph/?file=Authoring_Guide.md#augmenting-the-targlet-induced-target-platform

merks:
@ruspl-afed @gnl42 

This effectively removed 1/2 the content of the *.target.

I need to have a look at the category.xml to see which is included there that is kind of pointless...

merks:
@ruspl-afed @gnl42

 I assume you are okay with this change? I'd like to do further category.xml cleanup as a separate PR subsequent to this one.

gnl42:
This is all "black magic" to me.
I'll leave it to your expertise

merks:
Yes p2, PDE, and Tycho all seem to have slightly different interpretations on occasion, which is kind of frustrating, and then there is the actual OSGi wiring behavior at runtime, which is more then slightly different...

ruspl-afed:
@merks thank you for your effort
I cannot state that I understand why particular unit is required to be specified while others are not.
For example, we do use EMF in Mylyn, and this is why it was in the target, but of cause it _may be_ visible through other dependencies.
Observing the direction of changes I'm not sure if explicit dependency specification is still reasonable at all, may be we should just specify aggregated update site from SimRel and pray for luck.
I would love to learn today's best practices, since I have a number of other projects to support.

merks:
> I cannot state that I understand why particular unit is required to be specified while others are not.

The whole provided/required capability model is quite complex.  😞  And with so many package imports, sometimes it's surprising which bundle exports that package and maybe there is more than one "totally different" bundle doing that.

> For example, we do use EMF in Mylyn, and this is why it was in the target, but of cause it _may be_ visible through other dependencies.

In the end, everyone uses EMF because the e4 model uses it.   So Mylyn doesn't actually use more of EMF than the Platform does, unlike ECF where Mylyn uses `org.eclipse.ecf.provider.jmdns` but the Platform does not.

> Observing the direction of changes I'm not sure if explicit dependency specification is still reasonable at all, may be we should just specify aggregated update site from SimRel and pray for luck.

It is actually a good idea to pull dependencies from the originating projects from a n update site that changes with a frequency similar to SimRel.  So I would probably use something newer than this so that you immediately is EGit breaks something on which you depend:

https://download.eclipse.org/egit/updates-7.7

But not every project provides a nice "moving target" repository...

It's not actually a good idea to use the aggregated update site, because that site contains your own content, including the 3rd party bundles that you contributed in the past, so you might not notice that something in Orbit is newer, changed, or missing, because you might just pull it from something old you did in the past.

> I would love to learn today's best practices, since I have a number of other projects to support.

Let's merge this and then I will look at each update site we're using here?  Okay?

I actually did review the category.xml content it mostly makes sense what I see there.  Even stuff I though, "why does this need to be there" turned out to be a transitive dependency of another 3rd party bundle that you do use directly, so that looks quite clean and good to me.  Even for Lucene you use a large set of bundles than the Platform so including them is a good precaution.

ruspl-afed (approved):
LGTM
Let's see what's next :)

ruspl-afed:
> In the end, everyone uses EMF because the e4 model uses it

Well, may be not everyone. In Passage licensing engine relies on EMF to read encrypted "grants" and doesn't rely on E4. 
However, Passage includes optional support for E4 to check if one allowed to use an RCP product after E4 startup.

PR #1171, opened 2026-06-30
Use moving target repositories to follow latest upstream project changes

merks:


merks:
@ruspl-afed 

Here's a proposal for changing all the repositories to moving-target repositories that automatically track the latest upstream project changes.

Of course life is a trade-off.   A build might break just because of upstream changes; of course that's probably better than finding out the bundle doesn't work when SimRel produces M*/RC*.    One might use new API without noticing and forget to increment the lower bound; of course that probably already happens.  Not all these repositories are so nicely maintained using composites but rather overwrite the contents at the location so on rare occasions that could cause a glitch.

The only repository that continues to need maintaining is the ECF repository.  It remains like this because we redistribute ECF in our p2 repository, so don't want to redistribute a nightly (SNAPSHOT) build that might then end up in our release...

What do you think?

ruspl-afed:
> What do you think?

I don't know what to say. 
It definitely reduces the "M[ilestone]-ly" support effort at the cost of building from generally unpredictable content.

Please consider the following scenario: 
Platform produced an RC2 on Friday 
then version+1 effort was applied on next Monday or so with new I-build available
then on Wednesday we do Mylyn RC2 that grabs Platform with version+1 in a not yet released SimRel
WDYT?

merks:
> I don't know what to say. It definitely reduces the "M[ilestone]-ly" support effort at the cost of building from generally unpredictable content.
> 

It's all a trade-off and one can always paint a scenario where one way is better than another way, and for different reasons where one then needed to consider the prioritization of what's important.  Yes, the content might change in unpredictable ways, but those content changes will percolate though the system and will  pop out in SimRel.   At what point does Mylyn find a problem with the upstream content?  After it's been contributed to SimRel?  It's probably too late to do much about it then.  Do you find out the day after it happens?  It's probably easier to track down the cause and mitigate the problem; Oomph does nightly builds whether they are needed or not for this reason because I can't check out what might be broken on a daily basis myself.

> Please consider the following scenario: Platform produced an RC2 on Friday then version+1 effort was applied on next Monday or so with new I-build available then on Wednesday we do Mylyn RC2 that grabs Platform with version+1 in a not yet released SimRel WDYT?

It's a good question to ask when exactly does the following repository change!

https://download.eclipse.org/justj/?file=eclipse/updates/I-builds

I would ask Hannes, but he's on vacation.  But my guess its starts as soon as they start producing new I-build for the next cycle.  That's generally during the RC2 week for the rest of the projects.  In that case, it's a bad plan to use that.  The Oomph project setup updates the URL "at the right time" so it's probably best to use that URL in the catalog which is updated on the day of the release. 

https://github.com/eclipse-oomph/oomph/blob/dd115d123ffd5986e59d2f0a69b34ceff53bb4e4/setups/org.eclipse.projects.setup#L428-L433

The generator will update it to that new version the next time the setup tasks are performed...

So I would change this URL back to what it was.

Do you have other concerns? (This one was a good one!)

ruspl-afed (approved):
For the rest of URLs I do not have so clear concerns.
I'm very conservative, but I'll trust your expertise and experience here.
Yes, I understand that we already promoted Oomph to be a primary source of truth and we expect other project to play safe and according to SimRel rules (not always a case)

merks:
In the end, according to the build, the committed version of the *.target file is the source of truth.  One can always modify the *.target and commit the changes, one can activate it to use it in the IDE, and I left the old tasks disabled in the setup:

image

So yes, I think generating the target needed for the actual dependencies of the clone's project's is better, but it's just a tool and nothing is cast in stone.

---

Many folks don't know that Tycho actually doesn't really need a *.target either.  It can also work pretty much just like the targlets do where you only list repositories and not the specific dependencies that you require from those repositories; those are resolved by Tycho as it builds the projects.

ruspl-afed:
> Many folks don't know that Tycho actually doesn't really need a *.target either. It can also work pretty much just like the targlets do where you only list repositories and not the specific dependencies that you require from those repositories; those are resolved by Tycho as it builds the projects.

Interesting. However, not yet clear how we can use this fact to simplify our lives. For me the "simplicity" here is "an amount of sequenced steps that one need to perform to change dependency location/version"
Ideally, it should be a single change in one textual manifest, propagated to other places with CLI tooling. Concrete manifest does not matter for me, much more important to not repeat the same data in several places manually, because such kind of data duplications is the primary source of releng errors.

merks:
Suppose I decided to just jackson and so browsed around and decided I'll need to import this package:

image

So I textually edit the MANIFEST.MF to add it which has errors because it's not in the target platform.

So I perform the targlet task,.  Now the *.target is regenerated with the additional dependency:

image

So only one place needs to define it.  And, if I stopped using that dependency, the generator would remove it too.

Oh, and I forgot, Oomph contributes Discover... buttons to make it really easy to add things that are not already in your target.

image

Issue #1176, opened 2026-07-08
remove `org.apache.commons.lang3` from `org.eclipse.mylyn.wikitext.asciidoc`


ruspl-afed:
currently we use only 2 utility methods about nullable `String` handling

this doesn't worth constant effort to support 3rd party dependency

PR #1177, opened 2026-07-08
[#1176] remove org.apache.commons.lang3

ruspl-afed:
* replace with `Optional`


merks:
Oops, I meant to just make the windows job rerun.  Sorry

merks:
There are quite some uses elsewhere:

image

ruspl-afed:
> There are quite some uses elsewhere:

right, my current work is focused to let @azoitl consume less complicated dependency from Mylyn

merks:
It builds cleanly this time around.  :-)

azoitl:
@ruspl-afed that is very nice of you. But in the end I hope that any work that I do or trigger will make Eclipse Mylyn as a whole better in the end.

PR #1179, opened 2026-07-08
follow up for #1149 (reuse wikitext in 4diac)

ruspl-afed:
see #1149
* make `org.eclipse.mylyn.wikitext.asciidoc` accessible
* make `org.eclipse.mylyn.wikitext.html` accessible

### DIFF
--- a/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/META-INF/MANIFEST.MF
+++ b/mylyn.docs/wikitext/core/org.eclipse.mylyn.wikitext.asciidoc/META-INF/MANIFEST.MF
@@ -15,7 +15,7 @@ Import-Package: org.eclipse.mylyn.wikitext.parser;version="[4.2,5)",
  org.eclipse.mylyn.wikitext.parser.markup.token;version="[4.2,5)",
  org.eclipse.mylyn.wikitext.parser.outline;version="[4.2,5)",
  org.eclipse.mylyn.wikitext.util;version="[4.2,5)"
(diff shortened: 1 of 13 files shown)


merks:
FYI, I see API errors like this in my workspace:

image

azoitl:
@ruspl-afed I think this is more then what I and any consumer of Wikitext would need. Some service for creating the parser (Asciidoc, HMTL) would be sufficient and I think.

ruspl-afed:
> FYI, I see API errors like this in my workspace:

Thanks, fixed. It was re-promoted to an API since 4.12

ruspl-afed:
> I think this is more then what I and any consumer of Wikitext would need

This API is a kind of "package deal": it is hard to promote one part without another. 
We have `org.eclipse.mylyn.wikitext.ui.WikiText` that can give one an instance of `MarkupLanguage` by String `name` that will require further downcast, and I don't believe that it`s better than just instantiate corresponding "language".
TBH, I tried to redesign all this story to a more modern shape, but realized that it is above my current resourcing.
So, let's use it as it was designed many years ago, until we will see an opportunity to create something better.

azoitl:
> > I think this is more then what I and any consumer of Wikitext would need
> 
> This API is a kind of "package deal": it is hard to promote one part without another. We have `org.eclipse.mylyn.wikitext.ui.WikiText` that can give one an instance of `MarkupLanguage` by String `name` that will require further downcast, and I don't believe that it`s better than just instantiate corresponding "language". TBH, I tried to redesign all this story to a more modern shape, but realized that it is above my current resourcing. So, let's use it as it was designed many years ago, until we will see an opportunity to create something better.

I think a user like Eclipse 4diac could use it without the down cast. I currently have one in my code because it was easier but that is in general a bad idea.

PR #1180, opened 2026-07-08
[#567] remove gerrit dependencies

ruspl-afed:
* use what Tool produced


merks (approved):
Yes, that looks like the latest generator hard at work.  

Thanks so much for your dedication to the project.  

Also @gnl42 for all the cleanups and improvements!

PR #1186, opened 2026-07-09
Adjust unit test "timeout" value based on load average #1185

### LINKED ISSUE #1185: System load sensitive unit test
Failed twice on Mac:
```
org.eclipse.mylyn.commons.activity.tests.AllActivityTests.testInactivityTimeout -- Time elapsed: 0.071 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected less than 5 < activeTime < 40, got 62 (slept 62 ms) ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:232)
	at org.eclipse.mylyn.commons.activity.tests.MonitorUserActivityJobTest.testInactivityTimeout(MonitorUserActivityJobTest.java:61)
```

gnl42:
Test failed twice in MacOS. See if adjusting expected time based on load average helps

Task-Url: https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1185

### DIFF
--- a/mylyn.commons/org.eclipse.mylyn.commons.activity.tests/META-INF/MANIFEST.MF
+++ b/mylyn.commons/org.eclipse.mylyn.commons.activity.tests/META-INF/MANIFEST.MF
@@ -10,6 +10,7 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="0.0.0",
  org.eclipse.mylyn.commons.sdk.util.junit5;bundle-version="4.12.0"
 Export-Package: org.eclipse.mylyn.commons.activity.tests;x-internal:=true
 Import-Package: org.apiguardian.api;version="[1.1.0,2.0.0)",
+ org.assertj.core.api;version="[3.27.0,4.0.0)",
  org.junit.jupiter.api;version="[6.0.0,7.0.0)",
  org.junit.jupiter.api.condition;version="[6.0.0,7.0.0)",
  org.junit.jupiter.api.extension;version="[6.0.0,7.0.0)",
--- a/mylyn.commons/org.eclipse.mylyn.commons.activity.tests/src/org/eclipse/mylyn/commons/activity/tests/MonitorUserActivityJobTest.java
+++ b/mylyn.commons/org.eclipse.mylyn.commons.activity.tests/src/org/eclipse/mylyn/commons/activity/tests/MonitorUserActivityJobTest.java
@@ -13,9 +13,13 @@
 
 package org.eclipse.mylyn.commons.activity.tests;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
 
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
@@ -44,6 +48,11 @@ public class MonitorUserActivityJobTest {
 
 	@Test
 	public void testInactivityTimeout() throws Exception {
+		// Unstable on Mac, maybe this will help
+		OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
+		double loadAvg = osBean.getSystemLoadAverage();
+		long expectedActiveTime = (long) (40 * (loadAvg < 0.0 ? 1.0 : loadAvg > 1.0 ? loadAvg : 1.0));
+
 		callback.lastEventTime = System.currentTimeMillis() - 201;
 		job.setInactivityTimeout(200);
 		job.run();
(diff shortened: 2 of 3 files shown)


org.eclipse.mylyn.releng/oomph/Mylyn.setup
```diff
@@ -425,6 +425,8 @@
             name="org.eclipse.mylyn.wikitext.extras.feature.feature.group"/>
         <requirement
             name="org.eclipse.mylyn.wikitext.feature.feature.group"/>
+        <requirement
```
merks: I doubt you need this.  

I don't have time to look at this until later today.

merks: I had time to test this now.  

I removed this change setup change.   Then I removed the assetj-core from the *.target target:

Image

Then I ran the Modular Targlet task and it add that unit back into the *.target.

It really should generally be sufficient to mention a bundle or package in a MANIFEST.MF and run the targlet task; you should not need to "redundantly" specify that requirement in the setup or anywhere else.

Don't forget about these two button which will let you add dependencies even before they are resolved into your target platform:

Image

So you can add things, and then run the targlet task which should add them to the IDE's target platform and into the *.target...

So please remove this change form the commit.

merks (dismissed):
Please remove the setup change and otherwise it looks great.

org.eclipse.mylyn.releng/oomph/Mylyn.setup
gnl42: Not sure what steps I ended up taking in the end.

I did add the package to the manifest and ran the setup task. Eclipse restarted but didn't update the .target so I used the Repository Explorer to add the package to the setup file and it worked.

But I just tried it again without the setup and this time setup did update the .target file

merks: When you do perform setup tasks, this is generally what shows up:

image

This is the most convenient way to trigger that:

image

The first is the p2 task.  This will install new updates for the IDE much like Help -> Check for Updates.  Most often you don't need that.  Mostly there will only be updates after an M*, RC*, or release.  A restart is required if there are updates.

The second and third tasks are for the API baseline.  You only need to do this once per release cycle at the start of the cycle.  And then while it updates it often seems to be not visible in the IDE's projects.  This is good therapy for that problem:

image

(Maybe one day I'll figure out how to force API tools to really recognize the changes.)

The fourth tasks is some Maven thing.  Goodness knows why it's needed.  It seems to run on each startup too.  I don't think you need to run this manually ever.

The last one is for the target platform.   This is the most common and important thing you need to do to update against dependencies and to update the target platform if you add/change/remove dependencies.  You can double click it such that all the other tasks are unchecked and it's the only checked task.  That will be your most common thing to do.  All the other tasks are not as important and not as frequently needed.

gnl42: @merks  How would I go about adding org.eclipse.jst.servlet.ui from https://download.eclipse.org/webtools/repository/latest?

merks: May I ask why you want to add such a dependency?

Currently Mylyn only depends on these yellow projects:

image

And not to slag WTP, but it's my biggest concern for SimRel because it's a challenge to get this fixed even with PRs. 

Of the small number of open issue in SimRel

https://github.com/eclipse-simrel/simrel.build/issues

Three of those four are due to WTP not taking the steps needed to eliminate them.  

So making changes so that Mylyn is downstream from that is a concern to me.

Furthermore, Mylyn is in every EPP product:

image

But jst is not:

image

Just looking at the repository, hmm, two versions:

image

It has quite some list of dependencies:

image

As such I'm very concerned what will all be dragged into every EPP product...

Can you elaborate the purpose?

gnl42: JavaStructureBridge currently has this:
```
	/**
	 * TODO: remove after WTP 1.5.1 is generally available
	 */
	private String getWtpElementHandle(Object object) {
		Class<?> objectClass = object.getClass();
		try {
			Method getProjectMethod = objectClass.getMethod("getProject"); //$NON-NLS-1$
			Object javaProject = getProjectMethod.invoke(object);
			if (javaProject instanceof IJavaProject) {
				return ((IJavaElement) javaProject).getHandleIdentifier();
			}
		} catch (Exception e) {
			// ignore
			e.printStackTrace();
		}
		return null;
	}

	private boolean isWtpClass(Object object) {
		try {
			return object != null && object.getClass().getSimpleName().equals("CompressedJavaProject"); //$NON-NLS-1$
		} catch (Throwable t) {
			// could have malformed name, see bug 165065
			return false;
		}
	}
```

It's a dubious change since there are no unit tests for this case so I have no idea how to test the code

merks: I found this:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/tests/org.eclipse.ui.tests.navigator/src/org/eclipse/ui/tests/navigator/jst/CompressedJavaProject.java

And this is the class:

https://github.com/eclipse-jeetools/webtools.javaee/blob/master/plugins/org.eclipse.jst.servlet.ui/servlet_ui/org/eclipse/jst/servlet/ui/internal/navigator/CompressedJavaProject.java

So the code is calling getProject reflectively.

That class implements IAdaptable (maybe it didn't in the past) but given that's the case, it dead code because it won't get past this else if branch:

image

So definitely lets not complicate things to test something that appears to be dead.

gnl42: Sounds good to me

merks: So it's probably better to delete that code. If the adaptable implementation doesn't work, better if jst implement the support for that; maybe it already does.  In any case, I assume no one has complained.  😁

ruspl-afed (approved):
LGTM
I suppose that `.target` change is what one will see after fresh install. Am I right @gnl42 ?

merks:
> LGTM I suppose that `.target` change is what one will see after fresh install. Am I right @gnl42 ?

Yes this is the generated update resulting from the additional new dependency in the MANIFEST.MF also in this PR.   I tested that aspect…

PR #1187, opened 2026-07-10
Remove old Orbit repository from the targlet's repository list

merks:


merks:
This has no build impact, so merging immediately.

ruspl-afed:
Thank you @merks for this cleanup!

PR #1218, opened 2026-08-13
PoC: Use hub4j/github-api instead of egit.github

### LINKED ISSUE #1217: Use github-api
Replace the egit github client with the hub4j/github-api

gnl42:
Proof of Concept replacing the egit github client with the github-api 2.x client

Works, but doesn't install in an eclipse installation. 

Using the 2.0-rc.7 maven release. Don't know when the 2.0 release will be available


merks:
That project seems to have been working on release candidates for a very long time (using an unusual version scheme).

image

Is a 2.x  RC necessary versus the 1.399 release?

Wrapping it yourself is not really okay for the purpose of actually delivering this::

<location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" missingManifest="generate" type="Maven">

I see this

image

Should the tests pass?

gnl42:
> That project seems to have been working on release candidates for a very long time (using an unusual version scheme).

True, but the code is (slowly) being worked on

> 
> Is a 2.x RC necessary versus the 1.399 release?

I figured there was a bigger chance of seeing the new github api supported with the 2.x branch

> 
> Wrapping it yourself is not really okay for the purpose of actually delivering this:

Never the intent to release it this  way. Would request a proper orbit once 2.0 released

> 
> Should the tests pass?
??? The egit tests were deleted

merks:
I see.  The orbit thing was working for the 1.x so I thought probably it will work fine for 2.x when it's released, but I'll need to look at the dependencies again:

> Missing requirement: org.kohsuke.github-api 2.0.0.v20260813-1000 requires 'java.package; com.fasterxml.jackson.datatype.jsr310 [2.21.0,3.0.0)' but it could not be found

We do already have a whack of other jackson things, but who knows yet what else it might need.

Issue #1229, opened 2026-08-26
Centralize `DocumentBuilderFactory` creation


ruspl-afed:
Provide pre-configured safe instance of `DocumentBuilderFactory` for further usage

merks:
Thank you 🙏 for working on this!

PR #1230, opened 2026-08-26
[#1229] Centralize DocumentBuilderFactory creation

ruspl-afed:
* enhance `CoreUtil` and use it instead of in-place creation


merks (approved):
This looks really good. It eliminates duplication and makes things simpler and more consistent.

Issue #44, opened 2023-01-29
Support Dark Theme


gnl42:
Mylyn doesn't play well with Dark Theme. 
It uses hard coded colors that are close to impossible to read in Dark Mode.

An example is TaskScalingHyperlink which uses CommonColors.HYPERLINK_WIDGET

`public class CommonColors {

        public static final Color HYPERLINK_WIDGET = new Color(Display.getDefault(), 12, 81, 172);

        public static final Color TEXT_QUOTED = new Color(Display.getDefault(), 38, 86, 145);
`

BeckerFrank:
What is with the other Color Definitions?
We also have Class org.eclipse.mylyn.internal.context.ui.ColorMap.

gnl42:
And org.eclipse.mylyn.internal.commons.ui.E4ThemeColor which tries to parse `#org-eclipse-ui-workbench-DARK_BACKGROUND` as a RGB value:

`java.lang.NumberFormatException: For input string: "or" under radix 16
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:668)
	at org.eclipse.mylyn.internal.commons.ui.E4ThemeColor.getRGBFromCssString(E4ThemeColor.java:57)
	at org.eclipse.mylyn.commons.ui.GradientColors.getRGBFromTheme(GradientColors.java:211)`

gnl42:
I could use suggestions/pointers on how to go about doing this.

I have no experience with SWT UI programming. Trying to make sense of the various articles on Eclipse themes but unsure how to actually apply them..

Are custom colours really needed? Or are Eclipse defaults good enough?

wimjongman:
> Are custom colours really needed? Or are Eclipse defaults good enough?

When possible, Eclipse defaults would be the way to go.

https://www.vogella.com/tutorials/Eclipse4CSS/article.html contains many clues, but you probably already found it.

Theming is based on CSS, and when the plugin spy is installed, you can use CTRL+ALT+SHIFT+F6 to see how the component is styled.

Please ask specific questions and I will help.

gruscoe-papajohns:
I am also facing this issue - and with the very latest release - Do we have any css info that can be manually added to 

`plugins/org.eclipse.ui.themes_1.2.2400.v20240213-1133/css/e4-dark_linux.css`

to specify font colors for mylyn?

[...]

PR #1231, opened 2026-08-28
Replace GIF icons with SVG equivalents #44

vogella:
GIF only supports 1-bit transparency, so anti-aliased icon edges were flattened against a light background when the icons were authored. On a dark theme that baked-in fringe shows up as a bright halo, and because the icons are fixed-size rasters they also look blocky on HiDPI displays. This replaces 36 of them with the SVG equivalents that already exist in the Eclipse platform, JDT and PDE repositories, so they scale and pick up the surrounding theme.

Every candidate was rendered side by side against the original before being applied. Matches that shared a filename but turned out to be a different glyph were deliberately left out, among them `resource_obj` (upstream is a clock), `focus` (a magnifier), `remove` (a red X that would collide with `delete`) and the `notification-close` pair (both resolve to the same grey X, which would lose the active/inactive distinction).

Roughly 250 referenced GIFs remain in the repository. Those are Mylyn-specific artwork with no upstream counterpart, most of it in `commons.ui`, and they need hand-authored SVGs rather than a copy.

GIF to SVG comparison

Each cell above shows `[ old GIF | new SVG ]` on white, then the same pair on the dark theme background.


vogella:
I assume the license vetting failure is not related to this PR

azoitl:
I don't know what the policy is with images and dependencies but I see here quite some images duplicated from Eclipse Platform, Eclipse JDT and other dependencies of Eclipse Mylyn. Therefore, my maybe stupid question, would it be possible to use the images from the original place to avoid duplication. I think this could save some memory. Most probably not much. But what I see more important with the work currently going on on providing a new icon theme these duplicated images would need replacement here as well.

merks:
The failure is network in nature

> [ERROR] Failed to execute goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check (default-cli) on project org.eclipse.mylyn-aggregator: Execution default-cli of goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check failed: java.net.http.HttpTimeoutException: request timed out -> [Help 1]

Reusing images would be nice, but it makes them API if we plan to reuse them.  E.g., org.eclipse.ui.ISharedImages.

merks:
I'm not sure RC2 is a good point in the release to replace icons...

vogella:
> I don't know what the policy is with images and dependencies but I see here quite some images duplicated from Eclipse Platform, Eclipse JDT and other dependencies of Eclipse Mylyn. Therefore, my maybe stupid question, would it be possible to use the images from the original place to avoid duplication. I think this could save some memory. Most probably not much. But what I see more important with the work currently going on on providing a new icon theme these duplicated images would need replacement here as well.

Well written SVG file are very small.

vogella:
> The failure is network in nature
> 
> > [ERROR] Failed to execute goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check (default-cli) on project org.eclipse.mylyn-aggregator: Execution default-cli of goal org.eclipse.dash:license-tool-plugin:1.1.1-SNAPSHOT:license-check failed: java.net.http.HttpTimeoutException: request timed out -> [Help 1]
> 
> Reusing images would be nice, but it makes them API if we plan to reuse them. E.g., org.eclipse.ui.ISharedImages.

I think you can still refer to images in other plug-in via the plugin:// notation even without making them API, see my old blog https://vogella.com/blog/eclipse-rcp-plug-in-reading-an-icon-out-of-the-icon-directory-of-the-plug-in/

But I think we discourage that, during the transition from GIF to png we broken everyone how did this.

ruspl-afed:
@vogella thank you for your effort!

Before accepting this set of changes I need to ensure that URIs are still consistent through all the Mylyn codebase.
It takes time and even after this we can hardly take the risk of introducing "red square" just before the RC2.
So, I agree with @merks , it should wait a bit until the beginning of the next release cycle.

> herefore, my maybe stupid question, would it be possible to use the images from the original place to avoid duplication.

@azoitl this question is not stupid for sure since it is not so trivial. 
In general - yes, one can reference an image from another bundle by platform URI. But, since URI of bundle resource is not a part of API one can not rely on it safely. And this PR is a perfect example how such a relation may be broken: the change from PNG to SVG looks like a local cosmetic improvement, but it breaks image URIs. There are other ways to reuse images depending on the use case, but all of them heavily couple bundles in a price of saving relatively small amount of resources.
The [discussion](https://github.com/eclipse-pde/eclipse.pde/discussions/1315) on this issue has not been completed.

vogella:
> I'm not sure RC2 is a good point in the release to replace icons...

No strong option here, Mylyn has been like this since a while so a bit longer will not be to bad. I assume Mylyn user do not use the dark theme or are not very sensitive against pixel errors.  

I'm just currently building a set of custom themes https://github.com/vogellacompany/eclipse.themes for testing and improving Eclipse theming capabilities. During that I noticed that for example in Dracula (AFAIK IntelliJs most popular theme) Mylyn looks bad. https://github.com/vogellacompany/eclipse.themes/blob/main/docs/images/dracula.png

merks:
Yes, of course on can refer to things by their location, but that assumes that that file with that name will always exist at that location.  I.e., it assume the location is effectively API.  But I don't think the Platform makes no such promises; maybe that's evolving with the the SVG migration; I'm not sure.  In any case, when such location-based reference breaks, it's something one likely won't notice immediately.

vogella:
> @vogella thank you for your effort!

Happy to help, my friend. 

> Before accepting this set of changes I need to ensure that URIs are still consistent through all the Mylyn codebase. It takes time and even after this we can hardly take the risk of introducing "red square" just before the RC2. So, I agree with @merks , it should wait a bit until the beginning of the next release cycle.

I checked, so all URI should be updated, of course we cannot 100% sure. Please note that platform persists old image references so you need to start with -clean for views or close and open the view to see the new icon. If you prefer to handle that smoother, you could leave the GIF files in for a while and hope that users re-open the views in the time and do not leave them open for years in the same workspace. In platform we left the png files in for a while.

If that is desired, I can update the PR.

By the way (not in my PRs, existing): builds.ui/plugin.xml:209-210 already references platform:/plugin/org.eclipse.ui/icons/full/elcl16/stop.png if the platform converts stop.png to SVG, Mylyn gets the red square.

vogella:
> Yes, of course on can refer to things by their location, but that assumes that that file with that name will always exist at that location. I.e., it assume the location is effectively API. But I don't think the Platform makes no such promises; maybe that's evolving with the the SVG migration; I'm not sure. In any case, when such location-based reference breaks, it's something one likely won't notice immediately.

Correct, and I agree, that was exactly my point why I recommend not using references to platform.

azoitl:
Thx for the feedback and the detailed explanations. I get the problems with using file names and file paths into other plugins. This could only be solved by introducing some indirection like the org.eclipse.ui.ISharedImages. However that would  mean a lot of work. While I see every loaded file that can be spared can have a performance impact, even if the files are small. But consistent icons is for me a more important use case. But I think this here is not the right place to discuss this.

gnl42:
> > I'm not sure RC2 is a good point in the release to replace icons...
> 
> No strong option here, Mylyn has been like this since a while so a bit longer will not be to bad. I assume Mylyn user do not use the dark theme or are not very sensitive against pixel errors.
> 
> I'm just currently building a set of custom themes https://github.com/vogellacompany/eclipse.themes for testing and improving Eclipse theming capabilities. During that I noticed that for example in Dracula (AFAIK IntelliJs most popular theme) Mylyn looks bad. https://github.com/vogellacompany/eclipse.themes/blob/main/docs/images/dracula.png

Mylyn has real issues with dark theme. I've fixed a coupe of issues (I use dark theme), but it really needs proper theme support

vogella:
> Mylyn has real issues with dark theme. I've fixed a coupe of issues (I use dark theme), but it really needs proper theme support

I'm currently the primary developer of the CSS engine (big refactoring for 2026-09 with ~ 20 000 line of code), so I can help with the issues. The biggest offer are for me the icons at the moment. Once this is merged, point me to the remaining issues, I can try to solve them.

ruspl-afed:
> I'm currently the primary developer of the CSS engine (big refactoring for 2026-09 with ~ 20 000 line of code)

It sounds epic!

> Once this is merged

I hope to open master for 4.13 in a few days, if nothing special happens on my side.

ruspl-afed (approved):
LGTM
Thank you @vogella

PR #1232, opened 2026-08-28
Remove unreferenced GIF icons

vogella:
Removes 55 GIF icons that nothing in the repository points at. Each one was checked against every tracked text file, not just Java and plugin.xml, so references from properties files and markup are covered as well. Thirteen of them live in an `icons/archive` folder, and `task-list.gif` was left behind when the Task List view icon moved to SVG.

Icons that are loaded without a file extension were deliberately excluded. The EMF item providers in `reviews.edit` build their paths as `getImage("full/obj16/Comment")` and let the resource locator resolve the extension, so those 19 GIFs are live despite never appearing in the source as `*.gif`. No other bundle loads icons that way, and there is no dynamic icon-name construction anywhere in the tree.

This is independent of the GIF to SVG replacement in #1231 and touches a disjoint set of files.


vogella:
@ruspl-afed please review

ruspl-afed (approved):
LGTM (ignoring dash failure as not related)

Thank you @vogella

PR #1234, opened 2026-08-28
Reuse existing tasks.ui SVG icons in other bundles #44

vogella:
`org.eclipse.mylyn.tasks.ui` was already migrated to SVG, but several of the same icons still exist as GIF in other Mylyn bundles. Since GIF carries only 1-bit transparency, those copies show a light fringe against a dark theme and stay blocky on HiDPI, while the SVG next door does not. This replaces 17 of them with the SVG that is already in the repository, so nothing is taken from another project and the icons become consistent with the ones in the Task List.

The SVGs are the newer drawing rather than a straight conversion, so the affected icons do change appearance slightly. `focus` is the most visible of these and also the most widespread, at seven bundles. Everything else is the same shape with cleaner edges.

Two things were deliberately left out. `PERSON_LARGE` keeps its GIF, because the 48 pixel person icon is a paler, different drawing and the 16 pixel SVG is not a substitute for it. Seven further GIFs matching these names are unreferenced inside their own bundle and were skipped here; they are additional orphans that #1232 missed, because that PR matched file names across the whole repository and these were referenced only from a different bundle.

Independent of #1231 and #1232, and touches a disjoint set of files. Built locally with `mvn verify` over the twelve affected bundles that are in the aggregator reactor.

in-repo SVG reuse comparison

Each cell shows `[ old GIF | tasks.ui SVG ]` on white, then the same pair on the dark theme background.


vogella:
@ruspl-afed please review

ruspl-afed (approved):
LGTM
Thanks @vogella

PR #1235, opened 2026-08-28
Replace question.gif with SVG #44

vogella:
Replaces `CommonImages.QUESTION` with the SVG from `org.eclipse.help.ui/icons/etool16/help.svg`. The GIF carries a white disc baked into its 1-bit transparency, which shows as a bright halo on dark themes, and its 16 pixel raster is blocky on HiDPI.

Kept separate from #1231 for two reasons: the source file is named `help.svg` rather than `question.svg`, so a name-based match does not find it and it needs to be judged by eye; and this icon is highly visible, which makes it worth reviewing on its own rather than inside a batch of 36.

question icon before and after


vogella:
@ruspl-afed please review

vogella:
I assume the build errors are unrelated

merks:
Yes, until this is complete, the builds will fail:

https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/1244

Generally all masters builds are failing right now:

https://ci.eclipse.org/mylyn/job/GitHub-mylyn-Orga/job/org.eclipse.mylyn/job/main/

ruspl-afed (approved):
LGTM
Thanks @vogella

